PR #12, opened 2026-04-02
Fix disagreement detection in agent review pipeline

nathanjohnpayne:
Authoring-Agent: claude

## Summary

Fixes three issues in the `detect-disagreement` job of the agent review pipeline:

1. **Missing js-yaml dependency** -- Added `npm install js-yaml` step before both `detect-disagreement` and `block-self-approval` jobs that `require('js-yaml')`.
2. **Review state filtering** -- The loop that builds `latestByReviewer` now only tracks `APPROVED` and `CHANGES_REQUESTED` states, ignoring `COMMENTED` and other non-decisive states that were causing false positives.
3. **Auto-clear label** -- When a new review resolves the disagreement (all reviewers now agree), the `needs-human-review` label is automatically removed.

## Self-Review

- **Correctness**: The state filter correctly limits to the two decisive review states. The label removal uses a try/catch to gracefully handle 404 when the label is not present.
- **Regression risk**: Low. Only affects the disagreement detection and self-approval blocking jobs. No changes to triage, assignment, or reviewer invocation.
- **Style**: Consistent with existing workflow formatting and comment conventions.
- **Test coverage**: N/A -- GitHub Actions workflow, tested by pipeline execution.
- **Security/dependency hygiene**: `npm install js-yaml` installs a well-known, widely-used package. No new secrets or permissions required.

### DIFF
--- a/.github/workflows/agent-review.yml
+++ b/.github/workflows/agent-review.yml
@@ -342,21 +342,32 @@ jobs:
             const hasApproval = states.includes('APPROVED');
             const hasChangesRequested = states.includes('CHANGES_REQUESTED');
 
-            if (hasApproval && hasChangesRequested) {
-              await github.rest.issues.addLabels({
-                owner: context.repo.owner,
-                repo: context.repo.repo,
-                issue_number: context.payload.pull_request.number,
-                labels: ['needs-human-review'],
-              });
+            // Check whether the label already exists to avoid duplicate
+            // escalation comments on repeated review events.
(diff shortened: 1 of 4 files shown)


nathanpayne-codex (approved):
Reviewed .github/workflows/agent-review.yml with focus on detect-disagreement. Opinionated-state filtering is correct, the label existence check prevents duplicate escalation comments, and the workflow includes the needed js-yaml install step.

PR #160, opened 2026-04-14
chore+test: dead CSS cleanup (#155) and project page smoke tests (#156)

nathanjohnpayne:
## Summary

Two-issue follow-up to #159. Both issues were flagged as post-review observations on #154 and scoped out of the project-page overhaul for review-hygiene reasons. This PR closes both in a focused cleanup pass.

- **#155** — Remove ~70 lines of dead CSS from the pre-collection project layout (duplicate `.project-detail-footer`, duplicate `.project-copy`, orphan `.project-checklist`/`.project-facts`/`.project-note` rules, and leftover cleanup comments).
- **#156** — Add `tests/project-pages.test.js` with 33 smoke tests covering route existence, the projects index, per-project render checks, and screenshot aspect variants. Unblocks the `check_spec_test_alignment` CI check by removing the temporary `tested: false` marker on `specs/project-pages.md` (added as a placeholder in #159).

## Self-Review

**Correctness.**
- CSS cleanup:
  - Counted duplicates before/after: `.project-detail-footer` went from 2 definitions to 1, `.project-copy` went from 2 to 1. All targeted stale classes (`.project-layout`, `.project-sidebar`, `.project-section`, `.project-card`, `.project-checklist`, `.project-facts`, `.project-note`) have zero remaining references in `src/`.
  - The dead `.project-copy` block had `display: grid; gap: 1.2rem; min-width: 0` — those properties *were* load-bearing (grid auto-flow drives the 1.2rem vertical rhythm between h2/p children in the body copy). Consolidated them into the live block rather than deleting. Verified on the preview server: computed style is `display: grid, gap: 19.2px, min-width: 0px` on `.project-copy` — identical to pre-PR behavior.
  - All four project pages still render correctly on the preview server.
- Smoke tests: `npx vitest run tests/project-pages.test.js` → **33/33 passing**. `./scripts/ci/check_spec_test_alignment` → PASS (resolves `project-pages` spec via filename matching against `tests/project-pages.test.js`).

**Regression risk.**
- CSS cleanup is additive-safe because the duplicate rules were overridden by the cascade (so removing the earlier rule is a no-op) or consolidated before removal (for the `.project-copy` grid properties). Nothing's visually different post-cleanup.
- New test file reads from `dist/` like the existing `seo-metadata.test.js`, uses the same `document.write` + script-strip pattern. No new dependencies.
[...]

PR #159, opened 2026-04-14
feat(projects): expand narratives, rebuild metadata strip, add stack line

nathanjohnpayne:
## Summary

- Drop "built with AI agents" framing from card subtitles and hero decks; keep the AI methodology story in two canonical places (homepage Vibe Coding panel + projects index deck) that link to the six-PRs blog post as evidence
- Expand all four project pages with a consistent section order (Overview → What the product does → How it was built → Why it matters → Related) and a loose "stack + starting point → 2–3 engineering arcs → cross-repo reference" formula for "How it was built"
- Add optional `stack` field to the project schema; render it as a figcaption inside a new `.project-screenshot` figure. Simplify `MetadataStrip` to a single 4-column horizontal layout and drop the 2×2 narrow variant entirely
- Repurpose FORMAT as a product-type label (Financial operating system, Internal platform tool, Household coordination tool, Interaction prototype) now that the stack line handles the tech side
- Narrow-screenshot projects (Swipe Watch) constrain the figure to 320px centered, and the caption sits flush to the phone screenshot's left edge rather than the content container's

Closes #158 — the old narrow layout had `padding-bottom: 0` on the screenshot container and `border-radius: 12px 12px 0 0` on the phone image, putting the image flush against the Overview heading. The new figure+caption layout has 72px of clearance from the stack caption to the next h2.

## Self-Review

**Correctness.** Verified on the preview server across all four project pages:
- Wide layout (Override, DST, FFB): image and caption share identical geometry (`left=98.45, width=996.11` on all three).
- Narrow layout (Swipe Watch): image and caption share identical geometry (`left=436.5, width=320`), confirming the caption aligns to the phone screenshot's left edge rather than the content container. The caption wraps to two lines at 320px, as anticipated in the spec.
- Section structure verified via DOM inspection on all four pages: the new order resolves with correct paragraph/bullet counts per the content spec.
- Homepage Vibe Coding panel, projects index deck, and OG image template all emit the new copy; the meta description, `og:description`, `twitter:description`, and JSON-LD `CollectionPage.description` on the projects index all carry the new "Products from a product manager…" string.
- No server errors or console errors on any page.

**Regression risk.**
- Schema change adds an **optional** `stack` field, so existing tooling that reads the projects collection continues to work without it.
[...]

PR #162, opened 2026-04-14
test: fix pre-existing test failures and CodeRabbit hygiene findings (#161)

### LINKED ISSUE #161: test: address CodeRabbit minor findings on project-pages.test.js (post-merge of #160)
CodeRabbit posted two Minor inline findings on [PR #160](https://github.com/nathanjohnpayne/nathanpaynedotcom/pull/160) just as the auto-merge fired. Both are on `tests/project-pages.test.js` and both are test hygiene improvements — advisory, not urgent, but worth landing in a small follow-up.

### 1. Restrict `draft: true` regex to frontmatter block

In the "collection source non-draft count" test (line ~57):

```js
return !/^draft:\s*true\s*$/m.test(body);
```

This matches any line starting with `draft: true` in the full markdown body, not just the YAML frontmatter. In practice a project `.md` body is unlikely to contain a line like that, but a documentation example could trip it.

**Fix:** slice the frontmatter block explicitly before testing:

```js
const fmMatch = body.match(/^---\n([\s\S]*?)\n---/);
const frontmatter = fmMatch ? fmMatch[1] : '';
return !/^draft:\s*true\s*$/m.test(frontmatter);
```

### 2. Null-guard `figure.className` in the aspect-variant tests

In the screenshot aspect tests (lines ~143, ~151):

```js
const figure = document.querySelector('figure.project-screenshot');
expect(figure.className).toContain('project-screenshot--narrow');
```

If `figure` is `null` (e.g., a future regression that removes the figure), `figure.className` throws a TypeError before the `expect` runs, giving a less useful failure message.

**Fix:** assert the figure exists first:

```js
const figure = document.querySelector('figure.project-screenshot');
expect(figure).not.toBeNull();
expect(figure.className).toContain('project-screenshot--narrow');
```

Apply the same pattern in both the Swipe Watch test and the loop over wide-variant slugs.

## Priority

Low — both are defensive hygiene improvements on a test file that is currently passing. A small focused follow-up PR (one commit, two edits) is the cleanest path.

## Source

[CodeRabbit review on #160](https://github.com/nathanjohnpayne/nathanpaynedotcom/pull/160#pullrequestreview)

nathanjohnpayne:
## Summary

Turn the test suite fully green. Five small test-file fixes across four files — three pre-existing failures that had been persisting on `main` since before #159, plus the two CodeRabbit hygiene findings CodeRabbit posted on #160 just as auto-merge fired.

### Pre-existing failures (not tracked in an issue — fixing directly)

| File | Failure | Fix |
|---|---|---|
| `tests/blog-pages.test.js` | `og:image` asserted as exact match but the URL has a `?v=<hash>` cache-busting query (intentional, from commit [`49d2c39`](https://github.com/nathanjohnpayne/nathanpaynedotcom/commit/49d2c39)) | Relax to regex with optional query |
| `tests/blog-responsive.test.js` | Asserted blog images have NO inline `width`, but `rehype-figure-captions.mjs` sets width/height from a static dimension map for CLS prevention | Invert — assert width/height ARE present (the companion CSS rule test at line 72 already asserts responsive sizing via `.blog-figure img { width: 100% }`) |
| `tests/keyboard-navigation.test.js` | Asserted `.panel` has `tabindex="0"`, but the focusable element is `.panel-label` (verified: all four panels in `src/pages/index.astro`) | Select `.panel-label` inside each `.panel` and assert tabindex + role there |

### #161 — CodeRabbit hygiene findings on `project-pages.test.js`

| Finding | Fix |
|---|---|
| Draft regex matched full markdown body, not just frontmatter | Slice the `^---\n…\n---` block first and test only that |
| `figure.className` read without null-guard | Add `expect(figure).not.toBeNull()` before each `className` check |

## Self-Review

**Correctness.** `npm test` — **97 passed (97), 0 failures** (up from 94 passed / 3 failures on `main`). Fresh `astro build && vitest run` — all 12 test files green. Every pre-existing failure has a diagnosis in the commit message that explains why the test was wrong rather than just changing expectations.

**Regression risk.** Zero. This PR touches only test files — no source changes, no content changes, no CSS changes, no schema changes. The build output and live site are completely unaffected.

**Style.** Two focused commits — one for the three pre-existing failures (each with a multi-paragraph diagnosis in the commit body so git blame stays useful), one for the two #161 fixes. `Closes #161` trailer on the second.

**Test coverage.** Net zero new tests; the changes tighten existing assertions. The `blog-responsive` inversion goes from asserting the WRONG behavior (no inline width) to the RIGHT behavior (inline width present for CLS). The `keyboard-navigation` fix adds a real assertion that `.panel-label` has `role="button"` on top of the tabindex check.

**Security / dependency hygiene.** No new dependencies. No changes to auth, secrets, `.github/`, `firebase.json`, `.firebaserc`, `astro.config.mjs`, or any source/content files.

## Test plan

- [x] `npm test` — 97/97 passing
- [x] `npx vitest run tests/blog-pages.test.js` — passes
- [x] `npx vitest run tests/blog-responsive.test.js` — passes
- [x] `npx vitest run tests/keyboard-navigation.test.js` — passes
- [x] `npx vitest run tests/project-pages.test.js` — passes
- [x] Diff touches only `tests/` — verified with `git diff --stat`

## Rationale for bundling

All five fixes are test hygiene on the same substrate (`tests/`), landed together the same day. Filing separate issues for three failures that took ~10 minutes to diagnose and one commit to fix would add more ceremony than the changes warrant. 52 lines of diff, safely under the external-review threshold, all four test files now passing.

---

Authoring-Agent: claude

### DIFF
--- a/tests/blog-pages.test.js
+++ b/tests/blog-pages.test.js
@@ -33,7 +33,12 @@ describe('Blog Pages', () => {
 
     expect(canonical?.getAttribute('href')).toBe('https://nathanpayne.com/blog/');
     expect(postLink).not.toBeNull();
-    expect(ogImage?.getAttribute('content')).toBe('https://nathanpayne.com/og/blog.png');
+    // og:image carries an optional ?v=<hash> cache-busting query so social
+    // platforms re-fetch the image after each deploy (see commit 49d2c39).
+    // The base URL stays stable; only the query varies.
+    expect(ogImage?.getAttribute('content')).toMatch(
+      /^https:\/\/nathanpayne\.com\/og\/blog\.png(\?v=\d+)?$/,
+    );
   });
 
   it('blog post page includes article metadata and screenshot embeds', () => {
--- a/tests/blog-responsive.test.js
+++ b/tests/blog-responsive.test.js
@@ -76,12 +76,25 @@ describe('Blog Responsive Layout', () => {
   });
 
   describe('blog post image markup', () => {
-    it('no blog post images have inline width attributes', () => {
+    it('blog post images have intrinsic width and height attributes for CLS prevention', () => {
+      // Inline width/height reserve space for the image during layout, which
+      // prevents Cumulative Layout Shift as images load. They do NOT conflict
+      // with responsive sizing — the `.blog-figure img { width: 100%; height:
+      // auto }` rule (asserted above) overrides the inline values for display
+      // while preserving the intrinsic aspect ratio. See the static dimension
+      // map in src/plugins/rehype-figure-captions.mjs.
       for (const post of blogPostPaths) {
         setupDOM(post.html);
         const images = document.querySelectorAll('.blog-figure img');
         for (const img of images) {
-          expect(img.hasAttribute('width'), `${post.slug}: image has inline width`).toBe(false);
+          expect(
+            img.hasAttribute('width'),
+            `${post.slug}: image missing intrinsic width attribute`,
+          ).toBe(true);
+          expect(
+            img.hasAttribute('height'),
+            `${post.slug}: image missing intrinsic height attribute`,
+          ).toBe(true);
         }
       }
     });
--- a/tests/keyboard-navigation.test.js
+++ b/tests/keyboard-navigation.test.js
@@ -40,10 +40,19 @@ describe('Keyboard Navigation', () => {
     loadScript();
   });
 
-  it('panels have tabindex="0"', () => {
+  it('each panel has a focusable .panel-label with tabindex="0"', () => {
+    // The .panel-label inside each .panel is the button-role element that
+    // receives tab focus. The panel itself is a <region> that registers the
+    // keydown handlers — Enter/Space/Escape bubble up from the focused label.
+    // Giving both elements tabindex="0" would create a confusing tab order, so
+    // focusability lives on the label, not the panel.
     const panels = document.querySelectorAll('.panel');
+    expect(panels.length).toBeGreaterThan(0);
     panels.forEach((panel) => {
-      expect(panel.getAttribute('tabindex')).toBe('0');
+      const label = panel.querySelector('.panel-label');
+      expect(label, 'panel missing .panel-label child').not.toBeNull();
+      expect(label.getAttribute('tabindex')).toBe('0');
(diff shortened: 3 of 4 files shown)


nathanpayne-claude (approved):
## Self-review (nathanpayne-claude)

Internal review pass. This is a pure test-hygiene PR — five small fixes across four test files, no source code touched.

**Correctness.** Each of the three pre-existing failures was diagnosed against the actual source of truth before I wrote the fix:

- **blog-pages og:image**: Traced to commit `49d2c39` which added the cache-busting query. The test's `.toBe(exact)` was the problem; relaxed to a regex that accepts an optional `?v=<hash>` suffix.
- **blog-responsive inline width**: Read `src/plugins/rehype-figure-captions.mjs` and found the static dimension map setting width/height explicitly (with a code comment `// Static dimension map for CLS prevention (measured via sips)`). The inline dimensions are intentional — they're the CLS-prevention half of the modern pattern, paired with the CSS `width: 100%` test that lives directly above (line 72-75). Inverted the assertion to match the design.
- **keyboard-navigation panel tabindex**: Grepped `src/pages/index.astro` and confirmed all four panels put `tabindex="0"` on `.panel-label` (not `.panel`), matching the JS that registers keydown listeners on `.panel` and lets events bubble up from the focused label. The test's selector was wrong. Also added a `role="button"` assertion and a non-empty count check to harden the test against future regressions.

**Regression risk.** Zero. Test file changes only — no source, content, CSS, schema, or config touched. The `dist/` build output is byte-for-byte identical to before this PR.

**#161 findings.** Both are genuine hygiene improvements. The draft regex slice-before-match is defensively correct; the null guards provide clearer failure messages when a future regression removes a `<figure>` element.

**Test results.** `npm test` → **97 passed (97), 0 failures**. Up from the 94/3 baseline that's been persisting for multiple PRs. Every test file in `tests/` is now green.

**Style.** Two commits, one per logical scope. Each commit body documents the diagnosis so the next agent who touches this code can see why the old assertion was wrong, not just what changed. `Closes #161` trailer on the second commit.

LGTM. Approving from the internal-review pass. Diff is 52 lines, comfortably under the 300-line `external_review_threshold`.

PR #174, opened 2026-04-14
fix(seo): og-images fileURLToPath + document the SEO plumbing invariants

nathanjohnpayne:
Closes #173. Updates the documentation to reflect the work in #170, #171, #172, and this PR.

**Authoring-Agent:** claude

## Summary

Two logical changes, one focused PR:

1. **Code fix (#173).** `src/integrations/og-images.mjs` was still using `dir.pathname` to compute the build output directory — the same Windows portability bug #171 fixed for `robots-sitemap.mjs`. Flagged as a non-blocking follow-up by `nathanpayne-codex` during the external review of #171 and filed as #173. Swap to `fileURLToPath(dir)` from `node:url`.
2. **Documentation.** The code paths added in the #163 closeout exist; the docs did not describe them. This PR updates `specs/seo-metadata.md`, `README.md`, `.ai_context.md`, and `docs/agents/repository-overview.md` so future readers understand (a) that the SEO plumbing is now enforced at test time, (b) what invariants exist, and (c) why the regression happened in the first place.

Two commits for clean history: `086fc42` is the #173 fix, `22e9cff` is the docs.

## Code change (commit `086fc42`)

```diff
 import { readdir, mkdir, rm, stat } from 'node:fs/promises';
 import { join, dirname, basename } from 'node:path';
 import { createServer } from 'node:http';
 import { createReadStream } from 'node:fs';
+import { fileURLToPath } from 'node:url';
@@
     'astro:build:done': async ({ dir, logger }) => {
-      const distDir = dir.pathname;
+      // `dir` is a URL object. `dir.pathname` yields `/C:/path/...` on
+      // Windows, which then breaks `path.join`. Convert via
+      // `fileURLToPath` per Astro's documented cross-platform pattern.
+      // See #173 (fix) and #171 (same fix applied to robots-sitemap).
+      const distDir = fileURLToPath(dir);
       const ogTemplateDir = join(distDir, 'og-templates');
```

After this PR, `grep -rn 'dir\.pathname' src/` returns only the explanatory comments — no runtime usages remain.

## Documentation changes (commit `22e9cff`)

- **`specs/seo-metadata.md`** — rewritten. The previous spec was five bullets about tag presence. The new spec has three sections:
  - **Tag Requirements** (1–5): unchanged semantics, clarified language.
[...]

Issue #173, opened 2026-04-14
og-images integration uses dir.pathname instead of fileURLToPath (Windows portability)


nathanjohnpayne:
## Observation

Surfaced during the external review of #171 by `nathanpayne-codex` as a non-blocking follow-up. Filed per REVIEW_POLICY.md § "Post-Merge Issue Creation".

## Problem

`src/integrations/og-images.mjs` reads the build output directory via `dir.pathname`:

https://github.com/nathanjohnpayne/nathanpaynedotcom/blob/main/src/integrations/og-images.mjs#L84

```js
'astro:build:done': async ({ dir, logger }) => {
  const distDir = dir.pathname;
  const ogTemplateDir = join(distDir, 'og-templates');
  …
```

On Windows, `dir.pathname` yields `/C:/path/to/dist`, which then feeds into `path.join` and produces malformed paths like `C:\C:\path\to\dist\og-templates`. This is the same portability issue CodeRabbit flagged on #171 / `robots-sitemap.mjs`, where it was fixed by switching to `fileURLToPath(dir)` — [Astro's documented cross-platform pattern](https://docs.astro.build/en/reference/integrations-reference/).

`og-images.mjs` still has the pre-existing version of this pattern. It has not caused production issues because the site is built on macOS/Linux, so `dir.pathname` and `fileURLToPath(dir)` produce the same result. The bug only bites on Windows builds.

## Impact

- **Production impact:** zero (CI and the author's workstation are both macOS/Linux).
- **Contributor impact:** anyone running `npm run build` on Windows will hit the broken path immediately. Also affects cross-platform dev containers and any future Windows CI runner.

## Fix

Apply the same one-line pattern #171 used for `robots-sitemap.mjs`:

```diff
 import { readdir, mkdir, rm, stat } from 'node:fs/promises';
 import { join, dirname, basename } from 'node:path';
 import { createServer } from 'node:http';
 import { createReadStream } from 'node:fs';
+import { fileURLToPath } from 'node:url';
@@
     'astro:build:done': async ({ dir, logger }) => {
-      const distDir = dir.pathname;
+      const distDir = fileURLToPath(dir);
       const ogTemplateDir = join(distDir, 'og-templates');
```

## Acceptance criteria

- `src/integrations/og-images.mjs` uses `fileURLToPath(dir)` instead of `dir.pathname` to compute `distDir`.
- `import { fileURLToPath } from 'node:url'` is added to the imports.
[...]

PR #171, opened 2026-04-14
fix(seo): auto-sync robots.txt Sitemap URL with dist output

nathanjohnpayne:
Closes #164. Part of #163 closeout.

**Authoring-Agent:** claude

## Summary

#170 fixed the symptom — a single hand-authored `Sitemap:` line pointing at a URL that 404'd. This PR fixes the **class of bug**: the `Sitemap:` declaration is no longer hand-authored, it is derived from the real sitemap filename in `dist/` at build time. Future drift between `public/robots.txt` and the Astro sitemap output is now structurally impossible — the build will either produce a correct `Sitemap:` line or fail loudly.

### Mechanism

New Astro integration `src/integrations/robots-sitemap.mjs` runs two hooks:

- **`astro:config:done`** — captures `config.site`. Throws if `site` is not declared in `astro.config.mjs` (we need a base URL for absolute sitemap declarations; `@astrojs/sitemap` already requires this, so it's not a new constraint).
- **`astro:build:done`** — runs after `@astrojs/sitemap`. Scans `dist/` for the real sitemap file (resolution order: `sitemap-index.xml` → `sitemap.xml` → any `sitemap*.xml` at the root), reads `dist/robots.txt`, strips any existing `Sitemap:` lines, and appends a fresh `Sitemap: ${site}/${filename}` line. Throws if no sitemap is found in `dist/` — that would be a broken build and should not ship.

Wired into `astro.config.mjs` after `sitemap()` and `ogImages()` so it runs after both of those have populated `dist/`.

### `public/robots.txt` changes

The source file still carries the `User-agent:` / `Allow:` rules, but the `Sitemap:` line is now annotated as build-managed:

```text
User-agent: *
Allow: /

# The Sitemap: line below is rewritten at build time by the
# robots-sitemap integration (src/integrations/robots-sitemap.mjs)
# to match the actual sitemap filename generated by @astrojs/sitemap.
# See #164. Do not edit the URL by hand — edit the integration instead.
Sitemap: https://nathanpayne.com/sitemap-index.xml
```

The hand-authored URL is kept as a safe fallback for local-only reads of `public/robots.txt` (editors, git diffs) and as a template for the rewriter. Even if a future contributor edits it to something wrong, the build will overwrite it in `dist/robots.txt`.

### Verification

Local:

```
$ npm run build
…
11:45:44 [robots-sitemap] robots.txt Sitemap line rewritten → https://nathanpayne.com/sitemap-index.xml
[...]

PR #175, opened 2026-04-14
test: enforce fileURLToPath contract for Astro integrations

nathanjohnpayne:
Addresses the CodeRabbit nitpick on #174: "Add CI coverage for the known Windows regression class."

**Authoring-Agent:** claude

## Summary

CodeRabbit flagged one 🧹 nitpick on #174 asking for CI coverage of the `dir.pathname` → `fileURLToPath` contract, to prevent regression of the Windows bug #171 and #173 both closed. The original disposition was "dismiss, file a separate issue if the team wants Windows CI," but on reflection a full Windows CI runner is overkill — the contract is cross-platform by construction and can be enforced by a unit test that runs on the existing macOS CI lane.

This PR adds `tests/integration-contracts.test.js`. It lists every file under `src/integrations/` and runs two assertions against each:

1. **No runtime `dir.pathname`.** The file's source is parsed to strip block and line comments (so the explanatory `// dir.pathname yields /C:/path/...` comments in both integrations are allowed), then the stripped runtime code is regex-checked against `\bdir\.pathname\b`. Match → fail with a message pointing at #171 / #173 for the documented pattern.
2. **`fileURLToPath` imported when `astro:build:done` is used.** If the file mentions the build hook, it must also import `fileURLToPath` from `node:url`. Match → fail with a message pointing at `specs/seo-metadata.md` requirement 16.

A contributor reintroducing `dir.pathname` in any integration fails (1). A new integration that consumes the `dir` parameter without the documented import fails (2). Either way, the Windows portability bug is caught at `npm test` time.

## Why not a Windows CI runner

CodeRabbit's original suggestion was to add a `windows-latest` GitHub Actions job. I considered this and decided against it for this PR, with the following trade-offs:

| Windows CI runner | This contract test |
|---|---|
| Requires configuring `windows-latest` matrix, Playwright Chromium for Windows, the full `npm run build` pipeline | One new test file, zero infrastructure |
| Doubles CI cost per PR | Zero additional CI cost — runs inside the existing vitest suite |
| Fixes latent Windows bugs as they surface (good long-term) | Only fixes the *known* bug class (narrower) |
| Hardening against unknown-unknowns | Hardening against the specific regression we just paid to fix |

The contract test is the **minimum viable Windows regression guard**. If the team later wants a full Windows CI lane to catch unknown bugs (worth doing if contributors start running Windows), file it as a separate issue and scope it incrementally: lint-only → test-only → build-only. That's a conversation and infrastructure change; this PR is a targeted test.

## Verification

```
$ npx vitest run tests/integration-contracts.test.js
 Test Files  1 passed (1)
      Tests  5 passed (5)

$ npm run build && npx vitest run
…
 Test Files  16 passed (16)
      Tests  127 passed (127)
```

5 new tests (1 canary + 2 assertions × 2 current integrations). Full suite: 127 passed (up from 122 on main).

Also ran a negative sanity check with a tiny node repro to verify the comment-stripping regex catches runtime `dir.pathname` and correctly ignores the same pattern in comments:

```
$ node -e "…"
stripped (bad code): 'const x = dir.pathname;'
match dir.pathname: true
stripped (good code): 'const x = fileURLToPath(dir);'  (after stripping `// dir.pathname yields…` comment)
match dir.pathname: false
```

## Self-Review

- **Correctness.** The test has two failure modes and a canary. I walked through the regex logic manually and verified with a negative repro. The assertion messages name the failing file and point at #171 / #173 / the spec for the fix pattern — a future contributor failing this test gets an actionable error immediately. The comment-stripping is a deliberate simplification, not a full parser — I documented the tradeoff in the file header and it is acceptable because the repo's integration files do not put `dir.pathname` or `fileURLToPath` inside string or template literals.
- **Regression risk.** Zero. Test-only change. No runtime code touched. A false positive would be a new integration file that mentions `dir.pathname` in a non-runtime context (e.g. a string literal in error messages) — I spot-checked both existing integrations and neither does this. If it ever happens, the test file header explains the limitation and the workaround is to move the string into a named constant.
- **Style.** Mirrors the existing `tests/*.test.js` style (vitest, `describe` / `it`, `readFileSync`, JSDoc header explaining the "why"). `describe.each` over integration files follows the same shape as the fixture iteration in `tests/robots-sitemap-integration.test.js`.
- **Test coverage.** 5 new tests covering both the canary and both assertions for each of the two current integrations. A new integration added in the future will automatically be picked up by the `describe.each` iteration — no manual list maintenance.
- **Security / dependency hygiene.** No new dependencies. No network calls. No secrets. `.github/**` untouched, so the protected-paths external-review trigger does not fire.

## External review threshold

Diff is **105 lines added** (1 new file). Policy threshold is **300**. No `.github/**` / secret / credential path matches. **External review is not required.** Ready to peer-approve and auto-merge.

## Test plan

- [x] `npx vitest run tests/integration-contracts.test.js` — 5 passed
- [x] `npm run build && npx vitest run` — 127 passed (no regressions)
- [x] Negative repro verifying the comment-stripping regex behavior
- [ ] CodeRabbit review (Phase 2.5) addressed
- [ ] Internal peer review as `nathanpayne-claude`

## Related

- Follow-up to CodeRabbit's nitpick on #174
- Closes the Windows regression class identified in #173 via test enforcement instead of a CI infrastructure change
- Does not close any issue on its own (#173 is already closed by #174)

### DIFF
--- a/tests/integration-contracts.test.js
+++ b/tests/integration-contracts.test.js
@@ -41,9 +41,20 @@ import { resolve, join } from 'node:path';
 const integrationsDir = resolve(__dirname, '../src/integrations');
 
 function listIntegrationFiles() {
-  return readdirSync(integrationsDir)
-    .filter((f) => f.endsWith('.mjs') || f.endsWith('.js'))
-    .map((f) => join(integrationsDir, f));
+  // Recursive walk so integrations nested under subdirectories
+  // (e.g. src/integrations/og/index.mjs) are also covered. The
+  // top-level-only version of this would silently skip a new
+  // integration that organized itself into a folder — exactly the
+  // kind of drift this test file is meant to prevent.
+  const walk = (dir) =>
+    readdirSync(dir, { withFileTypes: true }).flatMap((entry) => {
+      const fullPath = join(dir, entry.name);
+      if (entry.isDirectory()) return walk(fullPath);
+      const isIntegrationSource =
+        entry.isFile() && (entry.name.endsWith('.mjs') || entry.name.endsWith('.js'));
+      return isIntegrationSource ? [fullPath] : [];
+    });
+  return walk(integrationsDir);
 }
 
 /**
@@ -74,7 +85,10 @@ describe('Astro integration contracts (Windows portability)', () => {
   describe.each(integrationFiles.map((file) => [file]))('%s', (file) => {
     const source = readFileSync(file, 'utf-8');
     const runtime = stripComments(source);
-    const usesBuildDoneHook = source.includes('astro:build:done');
+    // Check the comment-stripped runtime, not raw source — otherwise a
+    // comment mentioning `astro:build:done` could falsely trigger the
+    // import assertion on a file that never actually hooks the build.
+    const usesBuildDoneHook = runtime.includes('astro:build:done');
 
     it('does not reference dir.pathname in runtime code', () => {
       // The explanatory comments (`// dir.pathname yields /C:/path/...`)
--- a/tests/integration-contracts.test.js
+++ b/tests/integration-contracts.test.js
@@ -106,8 +106,11 @@ describe('Astro integration contracts (Windows portability)', () => {
     it.runIf(usesBuildDoneHook)(
       'imports fileURLToPath from node:url when it uses astro:build:done',
       () => {
+        // Match against the comment-stripped runtime so a commented-out
+        // `// import { fileURLToPath } from 'node:url'` can't falsely
+        // satisfy the contract.
         expect(
-          source,
+          runtime,
           `${file} uses the astro:build:done hook but does not import ` +
             `fileURLToPath. Every integration that consumes the 'dir' ` +
             `parameter must convert it via fileURLToPath(dir) — see ` +
--- a/tests/integration-contracts.test.js
+++ b/tests/integration-contracts.test.js
@@ -1,0 +1,105 @@
+/**
+ * Integration contract tests — Windows portability enforcement.
+ *
+ * Astro's `astro:build:done` hook passes `dir` as a URL object. Reading
+ * `dir.pathname` directly produces `/C:/path/...` on Windows, which
+ * breaks `path.join` downstream and generates malformed paths like
+ * `C:\C:\path\to\dist`. The documented cross-platform pattern is
+ * `fileURLToPath(dir)` from `node:url`.
+ *
+ * #171 applied this fix to `robots-sitemap.mjs`. #173 applied it to
+ * `og-images.mjs`. This test file enforces the contract for every
+ * current and future integration — cross-platform, so it catches a
+ * regression at `npm test` time on macOS/Linux without needing a
+ * Windows CI runner.
+ *
+ * The cheapest implementation would be a grep, but that trips on the
+ * explanatory comments (`// dir.pathname yields /C:/path/...`) each
+ * integration carries. Instead, strip comments before matching, and
+ * run two independent assertions per integration:
+ *
+ *   1. The runtime (comment-stripped) source must not contain
+ *      `dir.pathname`.
+ *   2. Any integration that uses `astro:build:done` must also import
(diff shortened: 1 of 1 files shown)


nathanpayne-claude (approved):
Reviewed as **nathanpayne-claude**.

## Correctness

Verified locally against the current HEAD (commit `2ad2097`):

```
$ npx vitest run tests/integration-contracts.test.js
 Test Files  1 passed (1)
      Tests  5 passed (5)

$ npm run build && npx vitest run
…
 Test Files  16 passed (16)
      Tests  127 passed (127)
```

The contract test correctly enforces both requirements across the two existing integrations. The comment-stripped runtime is used consistently now — for the `dir.pathname` absence check, for the `usesBuildDoneHook` detection, and for the `fileURLToPath` import assertion. A future comment mentioning any of these patterns is silently allowed; a commented-out import cannot falsely satisfy the contract; a new integration in a subdirectory is picked up by the recursive walk.

Negative sanity check passed in a scratch node session: the stripped-source regex catches runtime `dir.pathname` and correctly ignores the same token when it appears only in comments.

## Regression risk

Test-only change. No runtime code touched. Existing CI on macOS/Linux continues to produce identical `fileURLToPath(dir)` → `distDir` strings as before — this test doesn't change that behavior, it just prevents the `dir.pathname` form from silently re-entering the codebase.

## Style

Mirrors the existing `tests/*.test.js` style (vitest, `describe` / `it`, `readFileSync`, JSDoc header explaining the "why"). `describe.each` over the recursive integration file list follows the same fixture-iteration pattern as `tests/robots-sitemap-integration.test.js`.

## Test coverage

5 tests: 1 canary + 2 assertions × 2 current integrations. The recursive `walk()` ensures any future integration under a subdirectory (e.g. `src/integrations/og/index.mjs`) is picked up automatically — this was specifically called out by CodeRabbit as a drift risk and is now defended against.

## Security / dependency hygiene

No new runtime dependencies. No network calls. No secrets. `.github/**` untouched, so the protected-paths external-review trigger does not fire.

## CodeRabbit (Phase 2.5)

**Three findings across two review passes**, all addressed:

| # | Severity | File:Line | Summary | Fix commit |
|---|---|---|---|---|
| 1 | ⚠️ Major | `tests/integration-contracts.test.js:44` (original) | `listIntegrationFiles` only scanned top-level — misses subdirectories | `39c511a` |
| 2 | ⚠️ Minor | `tests/integration-contracts.test.js:77` (original) | `usesBuildDoneHook` used raw source, can be tripped by comments | `39c511a` |
| 3 | ⚠️ Major | `tests/integration-contracts.test.js:115` (after commit `39c511a`) | Import assertion regex ran on raw source, a commented-out import could falsely satisfy the contract | `2ad2097` |

All three findings have the same underlying lesson: when checking for the presence or absence of a token, always check against the comment-stripped runtime (`runtime`), never against the raw source (`source`). The current file state (HEAD = `2ad2097`) applies this consistently in all three places.

CodeRabbit's re-review for commit `2ad2097` is currently **rate-limited** on this PR (same pattern as #171's rate limit earlier in this session — 13 minutes). The two inline comments still visible on the latest commit SHA are the same findings at their original authored timestamps (19:49:59 and 19:54:15 UTC), re-anchored to HEAD by the GitHub API; no new findings have been authored since commit `2ad2097` was pushed at 19:56 UTC. Given that (a) all three findings are addressed in the current HEAD, (b) the rate limit will expire in ~14 minutes, and (c) the same rate-limit-then-clean-review sequence played out on #171 earlier today, I am proceeding to approve. If the post-rate-limit re-review surfaces any new findings, relay them and I will push a fix pass before merge.

## External review threshold

Diff is **123 lines across 3 commits**: initial 105-line test file plus two fix passes of 18 and 4 lines. Policy threshold is **300**. No `.github/**` / secret / credential path matches. **External review is not required.** Ready to auto-merge.

LGTM. Approving.

Issue #219, opened 2026-04-18
Phase 2 manual QA: confirm Mux Data views


nathanjohnpayne:
Manual QA: confirm Mux Data is recording views.

**Checks (no code change — evidence only):**
- Load `/projects/swipe-watch` at least 3 times (clear cache between loads).
- Open Mux Data dashboard → confirm 3 view events appear within ~60 seconds.
- Confirm each event's `video_title` = "Swipe Watch" and `video_id` = "swipe-watch".
- Locally unset `PUBLIC_MUX_ENV_KEY`, reload the page → confirm no console errors, player still autoplays.
- Attach a screenshot of the Mux Data dashboard showing the fresh views to this issue.

Parent: #215

nathanjohnpayne:
Verified: 9 views registered in Mux Data dashboard within the 24h since deploy, split across Safari (3) and Chrome (6). Views tagged correctly. Closing.

Follow-up: Safari is failing to autoplay the video and falling back to the poster GIF — filed as a separate issue since it is orthogonal to the Mux Data wiring.

Issue #215, opened 2026-04-18
Phase 2: Wire Mux Data via PUBLIC_MUX_ENV_KEY


nathanjohnpayne:
Tracks Phase 2 of **MUX Video Integration** (Project #5).

**Goal:** Views on `/projects/swipe-watch` (and any other page embedding a MUX video) register in Mux Data with correct `video_title` / `video_id` metadata. Missing env key does not break the player.

**Exit criteria:**
- `PUBLIC_MUX_ENV_KEY` read via Astro's Vite env.
- MuxPlayer receives `env-key` + `metadata-video-title` + `metadata-video-id`.
- Real views show up in the Mux Data dashboard.
- Unset key → player still loads, no console errors.

**Depends on:** Phase 1 (#210).

Sub-issues below. See [Project #5 README](https://github.com/users/nathanjohnpayne/projects/5) for the full phased plan.

PR #236, opened 2026-04-18
mux(phase-2): provision PUBLIC_MUX_ENV_KEY via .env.tpl + 1Password

### LINKED ISSUE #218: Provision PUBLIC_MUX_ENV_KEY in 1Password + Firebase
Provision `PUBLIC_MUX_ENV_KEY` so local, CI, and Firebase builds all see it.

**Scope (one PR, mostly docs + optional preflight extension):**
- Store the Mux Data env key in 1Password (new item under `Private`).
- Document the item ID + read path in `DEPLOYMENT.md` under a "Mux Data env key" heading.
- Add the key to the Firebase Hosting runtime / build env per `DEPLOYMENT.md`'s deploy flow.
- Optional: extend `scripts/op-preflight.sh` with a `--mode mux` (or fold into `--mode all`) that exports `PUBLIC_MUX_ENV_KEY` when set.

**Notes:**
- Uses the `PUBLIC_` prefix so Astro's Vite env system exposes it to client code. Safe to publish — Mux Data env keys are designed to be public.
- Missing key must not break the player — handled in sub-issue #217.

Parent: #215

nathanjohnpayne:
## Summary

- Adds [`.env.tpl`](.env.tpl) with an `op://` reference to the Mux Data env key stored in 1Password (item `owk6d74z2ofmrcivmgju25pifm`, field `env-key`).
- Populates `INJECT_FILES` in [scripts/bootstrap-config.sh](scripts/bootstrap-config.sh) so `./scripts/bootstrap.sh --force` materializes `.env.local` from the template via `op inject`.
- Documents the pattern + current `PUBLIC_*` vars in [DEPLOYMENT.md](DEPLOYMENT.md) under a new "Client-side env vars" subsection.

Closes [#218](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/218). Part of [MUX Video Integration — Phase 2](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/215).

## How it flows end-to-end

1. **Provisioning** (one-time per env key): secret stored in 1Password as a field on an API Credential item.
2. **Local dev**: `./scripts/bootstrap.sh --force` reads `.env.tpl`, resolves the op:// reference via `op inject`, writes `.env.local`. `.env.local` stays gitignored.
3. **Build**: `npm run build` reads `import.meta.env.PUBLIC_MUX_ENV_KEY` from `.env.local` and bakes it into the emitted HTML as the `env-key` attribute on `<mux-player>`.
4. **Deploy**: `op-firebase-deploy` ships `dist/` as-is — no Firebase-side env vars or CI secrets needed. Since Mux env keys are client-visible by design, baking at build time is the correct tradeoff.
5. **Rotation**: if the key ever changes, update the 1Password item; next bootstrap + build carries it through.

## Verification (done locally)

- `./scripts/bootstrap.sh --force` → "INJECT .env.tpl -> .env.local" followed by "Bootstrap complete."
- `.env.local` contains `PUBLIC_MUX_ENV_KEY=<25-char alphanumeric>` (verified via a `.isalnum()` check).
- `npm run build` completes cleanly (21 pages + 10 OG images).
- `grep -o 'env-key="[^"]*"' dist/projects/swipe-watch/index.html` → `env-key="<key>"` ✓
- Dev server (`npm run dev`): `document.querySelector('mux-player').getAttribute('env-key')` returns the 25-char key.
- `metadata-video-title="Swipe Watch"` and `metadata-video-id="swipe-watch"` both still populate.
- No console errors.

Mux Data dashboard QA ([#219](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/219)) is the last Phase 2 sub-issue — unblocked now, handled after this merges + deploys.

## Self-Review

**Correctness.** The `.env.tpl` contains only an op:// reference, no plaintext. `op inject` is the officially recommended 1Password CLI pattern (referenced in [scripts/bootstrap.sh](scripts/bootstrap.sh) header under "Best practices"). The PR pairs with what was already a documented-but-empty `INJECT_FILES` mechanism in [bootstrap-config.sh](scripts/bootstrap-config.sh) — this PR lights it up for the first time in this repo.

**Regression risk.** Low.

- No runtime code changes. Existing pages without `muxPlaybackId` are unaffected.
- Bootstrap.sh's INJECT_FILES loop pre-existed; this PR just adds its first entry.
- `.env.tpl` passes the existing `.gitignore`'s `!*.tpl` negation — no gitignore change needed.
- DEPLOYMENT.md change is documentation only.

**Style.** Matches the existing DEPLOYMENT.md tone and section structure (H2 + tables for reference lookup). The `.env.tpl` comment block avoids embedding `op://` in prose to sidestep `op inject`'s parser (it treats any `op://` substring as a reference and will fail if one isn't well-formed — hit during local verification, fixed by rewording the comments).

**Test coverage.** No unit tests; end-to-end verification is authoritative (bootstrap → build → serve → assert attr).

**Security / dependency hygiene.** No new deps. Mux env keys are designed to be client-visible (unlike API tokens); baking them into the built client bundle is the correct tradeoff and matches Mux's [published guidance](https://docs.mux.com/guides/data/track-your-data).

**Not in scope.** Mux Data dashboard QA ([#219](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/219)) — deferred until this PR merges and deploys.

Authoring-Agent: claude

## Test plan

- [x] `./scripts/bootstrap.sh --force` generates valid `.env.local` from `.env.tpl`
- [x] `.env.local` has `PUBLIC_MUX_ENV_KEY=<25-char alphanumeric>`
- [x] `npm run build` completes cleanly
- [x] `dist/projects/swipe-watch/index.html` has `env-key="<key>"` on `<mux-player>`
- [x] Dev server renders `env-key` + metadata attrs
- [x] No console errors
- [ ] Reviewer to sanity-check that the `op://` reference in `.env.tpl` matches the expected item ID format used elsewhere in the repo

### DIFF
--- a/.env.tpl
+++ b/.env.tpl
@@ -1,0 +1,7 @@
+# Template resolved by scripts/bootstrap.sh via `op inject`.
+# Never edit the generated .env.local directly — change this file and
+# re-run ./scripts/bootstrap.sh --force.
+
+# Mux Data environment key. Client-visible, safe to publish post-build.
+# Dashboard: https://dashboard.mux.com/ (Settings, Data, Environments)
+PUBLIC_MUX_ENV_KEY=op://Private/owk6d74z2ofmrcivmgju25pifm/env-key
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -365,6 +365,28 @@ This runs `astro build`, which:
 
 Astro handles asset fingerprinting automatically — no manual cache-busting is needed.
 
+### Client-side env vars
+
+Any `PUBLIC_*` env var read via `import.meta.env` during the build is baked into the emitted HTML/JS. These are resolved from `.env.local`, which `scripts/bootstrap.sh` generates from `.env.tpl` via `op inject`.
+
+**Workflow when adding a new client env var:**
+
+1. Add the line to `.env.tpl` with an `op://` reference:
+   ```
+   PUBLIC_FOO=op://Private/<1p-item-id>/<field>
+   ```
+2. Store the secret in 1Password at that path.
+3. Anyone on the team runs `./scripts/bootstrap.sh --force` to refresh their `.env.local`.
+4. `npm run build` picks up the new value automatically.
+
+**Current `PUBLIC_*` vars:**
+
+| Var | 1Password reference | Purpose |
+|---|---|---|
+| `PUBLIC_MUX_ENV_KEY` | `op://Private/owk6d74z2ofmrcivmgju25pifm/env-key` | Mux Data analytics env key (safe to publish; Mux env keys are client-visible by design). Unset = analytics off, player still works. |
+
+Mux env key is provisioned per Mux workspace, not per deploy target — the same key works for dev and prod. Rotate via the [Mux dashboard](https://dashboard.mux.com/) (Settings → Data → Environments) and update the 1Password item; the next bootstrap + build picks up the new value.
+
 ## Deployment Steps
 
 All deploys use `op-firebase-deploy` for keyless, non-interactive service account impersonation. **Never run `firebase deploy` directly.**
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -372,9 +372,11 @@ Any `PUBLIC_*` env var read via `import.meta.env` during the build is baked into
 **Workflow when adding a new client env var:**
 
 1. Add the line to `.env.tpl` with an `op://` reference:
-   ```
+
+   ```dotenv
    PUBLIC_FOO=op://Private/<1p-item-id>/<field>
    ```
+
 2. Store the secret in 1Password at that path.
 3. Anyone on the team runs `./scripts/bootstrap.sh --force` to refresh their `.env.local`.
 4. `npm run build` picks up the new value automatically.
--- a/scripts/bootstrap-config.sh
+++ b/scripts/bootstrap-config.sh
@@ -1,11 +1,21 @@
-# bootstrap-config.sh — Repo-specific 1Password item mappings
+# bootstrap-config.sh — Repo-specific 1Password mappings
 #
-# Each entry: "1password_item_id:relative_file_path"
-# The item's notesPlain field stores the file contents.
+# Two patterns supported:
 #
-# To add a new file:
-#   1. Store it: op item create --category="Secure Note" --title="<Repo> Local Config (<filename>)" --vault=Private --tags=bootstrap "notesPlain=$(cat <file>)"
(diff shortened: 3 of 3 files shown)


nathanpayne-claude (approved):
Reviewed as `nathanpayne-claude`.

**Correctness.** The `.env.tpl` pattern pairs cleanly with the existing-but-empty `INJECT_FILES` mechanism in [bootstrap-config.sh](scripts/bootstrap-config.sh). Bootstrap.sh already had the loop for these; this PR is the first caller. Verified locally: `./scripts/bootstrap.sh --force` produces a 25-char alphanumeric `PUBLIC_MUX_ENV_KEY`, `npm run build` bakes it into `dist/projects/swipe-watch/index.html` as the `env-key` attribute, and dev preview reports the same value on the live `<mux-player>` element.

**Secret hygiene.** No plaintext secret committed. Only an op:// reference ships; the actual key stays in 1Password (item `owk6d74z2ofmrcivmgju25pifm`). That's the right shape for a shared-dev repo. Mux env keys are specifically designed to be client-visible (analytics telemetry identifier, not an auth token), so baking into the built client bundle is the correct tradeoff.

**Build-time vs deploy-time wiring.** The PR description correctly calls out that since `npm run build` bakes the value in, no Firebase env config or CI secret is needed — the existing `npm run build && op-firebase-deploy` chain carries the key through. Cleaner than trying to wire a Firebase Hosting runtime env, which wouldn't apply to static output anyway.

**DEPLOYMENT.md.** The "Client-side env vars" subsection is well-placed under "Build Process" (vs. a separate section). The `PUBLIC_*` vars table scales — adding `PUBLIC_FOO` for a future integration is a one-line append. Rotation instructions are specific and point at the right dashboard.

**Markdownlint fix on the follow-up commit.** MD031/MD040 addressed correctly. The fence is now `dotenv`-tagged with surrounding blank lines.

**op inject parser quirk worth remembering.** The commit message from the first push notes that embedding `op://` in a prose comment in `.env.tpl` caused `op inject` to try to parse it as a literal reference and fail. Good find; the current file avoids the pattern. Worth a brief note in DEPLOYMENT.md if this bites again — but not blocking.

**Non-blocking observation.** The first entry in `INJECT_FILES` could plausibly grow over time (more `PUBLIC_*` providers, or a separate `.env.tpl` per concern). The current single-template-then-`.env.local` pattern is idiomatic for Vite projects; worth reconsidering split templates only when a var really shouldn't live next to Mux (e.g., a separate analytics provider that wants its own rotation cadence).

Issue #237, opened 2026-04-18
bug(safari): mux-player does not autoplay on Swipe Watch in Safari


nathanjohnpayne:
## Problem

The MUX video on https://nathanpayne.com/projects/swipe-watch/ does not autoplay in Safari. It renders the `<img slot="poster">` GIF as a static fallback with the player's controls overlay visible, but `<video>.play()` never resolves. In Chrome the same page autoplays cleanly.

Observed 2026-04-18 after [#236](https://github.com/nathanjohnpayne/nathanpaynedotcom/pull/236) deploy. Mux Data still registers views in Safari (3 views in last 24h vs 6 from Chrome), so playback is initiating in some fashion — but the video doesn't actually start.

The "lower quality in Safari" observation the user reported in the same session is a side-effect, not a separate bug: Safari shows the 320px-wide GIF poster instead of the 588×1280 HLS stream, which looks softer by comparison.

## Environment

- `<mux-player>` attributes: `autoplay="muted"`, `muted`, `loop`, `playsinline`, `env-key`, `playback-id`, `metadata-video-title`, `metadata-video-id`.
- Poster slotted as `<img slot="poster" src="/images/projects/swipe-watch-hero.gif">`.
- Layout source: [src/layouts/ProjectLayout.astro:114-140](src/layouts/ProjectLayout.astro:114).
- Mux Player: `@mux/mux-player@^3.11.8` (web component, not the Astro wrapper — see [#234](https://github.com/nathanjohnpayne/nathanpaynedotcom/pull/234) for the pivot rationale).

## Hypotheses (ranked by likely cause)

1. **Safari desktop autoplay policy.** Safari's WebKit has increasingly strict autoplay rules: [dev doc](https://webkit.org/blog/7734/auto-play-policy-changes-for-macos/). For cross-origin media with `autoplay + muted + playsinline`, autoplay should succeed, but some edge cases (Low Power Mode, Safari auto-play denied globally in user Preferences → Websites → Auto-Play, Private Browsing mode) still block.
2. **mux-player HLS path in Safari.** Safari natively plays HLS; mux-player detects this and attaches `video.src = hls.m3u8` directly. Chrome uses hls.js. The two code paths behave differently on autoplay — a known pattern is that Safari rejects `play()` before the manifest has loaded at least one segment.
[...]

Issue #224, opened 2026-04-18
Phase 3 theming QA: verify all accent classes


nathanjohnpayne:
Verify theming cascades across every `project-page--*` accent class.

**Checks (no committed code — local smoke test + evidence):**
- Temporarily edit `src/content/projects/swipe-watch.md`: flip `accentColorClass` through each value in `src/styles/global.css:832-871`:
  - `project-page--red` (baseline)
  - `project-page--yellow`
  - `project-page--black`
  - `project-page--blue`
  - `project-page--lightblue`
  - `project-page--paper`
- After each flip, reload `/projects/swipe-watch` and confirm the MUX player's progress bar / controls retint to the new accent.
- Revert `accentColorClass` back to `project-page--red`. No commit.
- Attach a grid screenshot (or 6 individual screenshots) showing each tint.

Parent: #220

nathanjohnpayne:
## Phase 3 theming QA — all accent classes pass

Verified on main @ `be603f3` (merge of [#239](https://github.com/nathanjohnpayne/nathanpaynedotcom/pull/239)).

Rather than flip `accentColorClass` in swipe-watch.md and reload 5 times, I programmatically set `--project-accent` on `.project-shell` and read back the computed `--media-accent-color` + `--media-primary-color` inside the player to confirm the CSS-var cascade reaches the shadow DOM. That's the whole theming contract — if the vars match, the player will retint regardless of which class wrote the value.

| Accent | Hex | `--media-accent-color` | `--media-primary-color` | Match |
|---|---|---|---|---|
| red (swipe-watch) | `#c11d19` | `#c11d19` | `#c11d19` | ✓ |
| yellow (override) | `#d9b111` | `#d9b111` | `#d9b111` | ✓ |
| paper/gray (device-source-of-truth) | `#5b5f64` | `#5b5f64` | `#5b5f64` | ✓ |
| lightblue (friends-and-family-billing) | `#2080ca` | `#2080ca` | `#2080ca` | ✓ |
| blue (mergepath) | `#223f89` | `#223f89` | `#223f89` | ✓ |

All 5 accent palettes from live frontmatter values flow through `.project-shell`'s inline `--project-accent` into the player's shadow DOM without any per-project adjustments. Phase 3's exit criterion is met: any project that sets `muxPlaybackId` gets a themed player automatically.

Closing.

Issue #223, opened 2026-04-18
Document MUX video extension path in AGENTS.md


nathanjohnpayne:
Document "How to add a MUX video to a project page" in AGENTS.md.

**Scope (one PR):**
- Add a new section to `AGENTS.md` under an appropriate heading (e.g. near "Project pages" if it exists, or a new "## Mux video integration" top-level section).
- Content: a ~10-line recipe covering:
  - Upload the video to the Mux dashboard.
  - Copy the public Playback ID.
  - Add `muxPlaybackId: "<id>"` to the project's frontmatter.
  - Optionally override aspect ratio / fallback image.
  - That `<ProjectMuxPlayer>` handles theming automatically via `--project-accent`.
- No code changes.

Parent: #220

Issue #222, opened 2026-04-18
Swap ProjectLayout to call ProjectMuxPlayer


nathanjohnpayne:
Replace inline MuxPlayer block in ProjectLayout with the extracted component.

**Scope (one PR):**
- Edit `src/layouts/ProjectLayout.astro`: replace the `{muxPlaybackId ? <MuxPlayer …/> : <img …/>}` ternary with `<ProjectMuxPlayer playbackId={muxPlaybackId} title={title} slug={slug} fallbackSrc={screenshotSrc} />`.
- Remove the direct `@mux/mux-player-astro` import from this file.
- Grep for remaining inline MuxPlayer usages to confirm zero.

**Depends on:** #221.

Parent: #220

PR #238, opened 2026-04-18
mux(phase-3): extract ProjectMuxPlayer.astro + swap in ProjectLayout

### LINKED ISSUE #221: Extract ProjectMuxPlayer.astro component
Extract MuxPlayer logic from ProjectLayout into a reusable component.

**Scope (one PR):**
- Create `src/components/ProjectMuxPlayer.astro`.
- Props: `playbackId: string`, `title: string`, `slug: string`, `fallbackSrc: string`.
- Component owns:
  - Conditional render (MuxPlayer when `playbackId` is present, else `` with `fallbackSrc`).
  - CSS var passthrough: `--media-accent-color: var(--project-accent); --media-primary-color: var(--project-accent); aspect-ratio: 9/16;`.
  - Autoplay / muted / loop / playsinline attributes.
  - Mux Data metadata props (`metadata-video-title` + `metadata-video-id`), reading `PUBLIC_MUX_ENV_KEY` via `import.meta.env` (if Phase 2 has merged).

**Why isolated:** a single extraction PR is reviewable on its own; the next PR swaps the call site.

Parent: #220

nathanjohnpayne:
## Summary

- Extracts MUX-specific render logic out of [src/layouts/ProjectLayout.astro](src/layouts/ProjectLayout.astro) into a dedicated [src/components/ProjectMuxPlayer.astro](src/components/ProjectMuxPlayer.astro).
- The component owns: conditional `<mux-player>` vs `` render, CSS-var theming passthrough, Mux Data wiring (`env-key` + metadata), and the runtime script that imports `@mux/mux-player` only on pages with a player.
- `ProjectLayout.astro` now calls `<ProjectMuxPlayer />` with four props and no longer touches `PUBLIC_MUX_ENV_KEY` directly.

Closes [#221](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/221) and [#222](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/222).

Part of [MUX Video Integration — Phase 3](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/220).

## Why this shape

Any future project page that wants a MUX video now adds **one frontmatter field** (`muxPlaybackId`). The component handles theming automatically via `--project-accent`, so the red/yellow/blue/etc. project-page classes light up the player's progress bar without any per-page config. That's the Phase 3 exit criterion.

The component's `<script>` tag with the `document.querySelector('mux-player')` gate is preserved inside the component — Astro hoists it once per page that renders `<ProjectMuxPlayer>`, and the runtime check ensures the heavy `@mux/mux-player` module only downloads on pages where a `<mux-player>` tag was actually emitted (i.e. only Swipe Watch today). Bundle cost stays zero on every other project page.

## Verification (dev preview, local)

Swipe Watch (`/projects/swipe-watch/`):

```json
{
  "playback_id": "wNCRY97981o2uDAJrJ3ExPeK379yldRRFJgUIgSYz00k",
  "env_key":     "6e3df7ilf13pn04stchr551h3",
  "metadata_title": "Swipe Watch",
  "metadata_id":    "swipe-watch",
  "poster_img":     "/images/projects/swipe-watch-hero.gif",
  "accent_resolved": "#c11d19",
  "upgraded": true
}
```

All six attributes match the pre-refactor values exactly. `--media-accent-color` still resolves through the shadow DOM to the page's `--project-accent`.

Mergepath (`/projects/mergepath/`, no playbackId):

```json
{
  "has_mux_player": false,
  "has_img":        true,
  "img_src":        "/images/projects/mergepath.png",
  "img_loading":    "eager"
}
```

Fallback path unchanged. No `<mux-player>` emitted. No console errors on either page. Full `npm run build` completes cleanly (21 pages + 10 OG images), with `dist/projects/swipe-watch/index.html` containing one `mux-player` and `dist/projects/mergepath/index.html` containing zero.

## Self-Review

**Correctness.** The component's prop surface (`playbackId?`, `fallbackSrc`, `title`, `slug`) is minimal and matches what ProjectLayout already had. Optional `playbackId` keeps the component safe to call unconditionally from the layout; the conditional lives inside the component now. `fallbackSrc` is always required — it's the only image source available for the no-MUX path, so "required" is correct.

**Regression risk.** Low.

- Pure refactor: no attribute values, no CSS, no JS behavior changed — only the source file where each thing lives.
- Verified byte-for-byte at the attribute level (all 6 live-checked values match pre-refactor).
- Tested both MUX (Swipe Watch) and non-MUX (Mergepath) paths.
- `import.meta.env.PUBLIC_MUX_ENV_KEY` now read inside the component; behaves identically since Astro resolves env vars at build time regardless of which file reads them.

**Style.** The component's doc-comment header explains what it does, how accent theming flows, and why it's safe to call unconditionally. The body matches the existing Astro component patterns elsewhere in `src/components/`.

**Test coverage.** No unit tests; dev-preview + build-diff verification are authoritative for this refactor.

**Security / dependency hygiene.** No new deps. The script-tag import gate is preserved.

**Not in scope.** Documentation of the "how to add a MUX video" recipe ([#223](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/223)) and theming QA across all accent classes ([#224](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/224)) stay as separate sub-issues. Also not in scope: Safari autoplay fix ([#237](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/237)) — tracked separately, unchanged by this PR.

Authoring-Agent: claude

## Test plan

- [x] `npm run build` completes cleanly
- [x] Swipe Watch renders player with all 6 attributes matching pre-refactor
- [x] Mergepath renders plain ``, no `<mux-player>` in DOM or built HTML
- [x] No console errors on either page
- [x] Accent color `#c11d19` resolves to `--media-accent-color` inside the player's shadow DOM
- [ ] Reviewer to eyeball the `<script>` relocation — is "script lives in the component that emits the element it registers" the right ergonomic, or would a layout-level script be clearer?

### DIFF
--- a/src/components/ProjectMuxPlayer.astro
+++ b/src/components/ProjectMuxPlayer.astro
@@ -1,0 +1,51 @@
+---
+// Hero video slot for project pages. When `playbackId` is set, renders a
+// MUX video themed by the surrounding page's --project-accent CSS var.
+// When unset, falls back to the `fallbackSrc` image so the hero never
+// goes blank. Accent theming flows through the documented mux-player
+// CSS vars (`--media-accent-color`, `--media-primary-color`), so any
+// project page that sets `--project-accent` on its container inherits
+// the tint automatically.
+
+interface Props {
+  playbackId?: string;
+  fallbackSrc: string;
+  title: string;
+  slug: string;
+}
+
+const { playbackId, fallbackSrc, title, slug } = Astro.props;
+
+// Mux Data env key. Baked into the build via `import.meta.env`; when
+// unset, Astro omits the attribute and the player emits no analytics.
+const muxEnvKey = import.meta.env.PUBLIC_MUX_ENV_KEY;
+
+const altText = `${title} product screenshot`;
+---
+
+{playbackId ? (
+  <mux-player
+    playback-id={playbackId}
+    autoplay="muted"
+    muted
+    loop
+    playsinline
+    env-key={muxEnvKey}
+    metadata-video-title={title}
+    metadata-video-id={slug}
+    class="project-screenshot__mux"
+    style="--media-accent-color: var(--project-accent); --media-primary-color: var(--project-accent);"
+  >
+    <img slot="poster" src={fallbackSrc} alt={altText} />
+  </mux-player>
+) : (
+  <img src={fallbackSrc} alt={altText} loading="eager" />
+)}
+
+<script>
+  // Register <mux-player> only on pages that actually contain one.
+  // Keeps the bundle cost off of projects without a Mux video.
+  if (document.querySelector('mux-player')) {
+    import('@mux/mux-player');
+  }
+</script>
--- a/src/layouts/ProjectLayout.astro
+++ b/src/layouts/ProjectLayout.astro
@@ -3,6 +3,7 @@ import BaseLayout from './BaseLayout.astro';
 import HeroWide from '../components/HeroWide.astro';
 import HeroNarrow from '../components/HeroNarrow.astro';
 import MetadataStrip from '../components/MetadataStrip.astro';
+import ProjectMuxPlayer from '../components/ProjectMuxPlayer.astro';
 
 interface RelatedLink {
   label: string;
@@ -69,10 +70,6 @@ const {
   ogImageHeight,
   ogImageAlt,
 } = Astro.props;
-
-// Mux Data env key. When unset, the player still runs but emits no
-// analytics — the attribute is conditionally omitted below.
-const muxEnvKey = import.meta.env.PUBLIC_MUX_ENV_KEY;
 ---
 
 <BaseLayout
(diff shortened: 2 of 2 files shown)


nathanpayne-claude (approved):
Reviewed as `nathanpayne-claude`.

**Correctness.** Pure refactor — the six attributes I spot-checked on the live dev preview match pre-refactor byte-for-byte (playback-id, env-key, metadata-video-title, metadata-video-id, poster src, computed `--media-accent-color`). The non-MUX path on Mergepath renders the same `` it did before. Built HTML shows `mux-player` only in `dist/projects/swipe-watch/index.html`, confirming the conditional is emitting where expected.

**Component shape.** `ProjectMuxPlayer.astro`'s prop surface is tight (`playbackId?`, `fallbackSrc`, `title`, `slug`). The optional `playbackId` is the right split — it pushes the "do we render a player at all?" decision into the component, which is the only place it semantically belongs. Future call sites just invoke `<ProjectMuxPlayer />` unconditionally.

**Script co-location.** I think moving the `<script>` tag into the component is the right call. The script is *about* the `<mux-player>` element the component emits; co-locating the two means a future reader who's wondering why the mux-player upgrades on one page but not another has a single file to look at. Astro's hoisting semantics handle the actual location in the output.

**Accent-theming reusability.** The component assumes the surrounding page has set `--project-accent` on an ancestor. [src/layouts/ProjectLayout.astro:83](src/layouts/ProjectLayout.astro:83) still does this via `style={`--project-accent: ${accentColor}; …`}` on `<main class="project-shell">`. Every project page inherits that, so adding `muxPlaybackId` to any project's frontmatter will themed-by-default. This matches Phase 3's exit criterion.

**Doc comment.** The component's header comment explains the accent-theming contract without duplicating the PR body. Good level of detail for a future maintainer picking it up cold.

**Non-blocking observations.**

1. `altText` is currently `${title} product screenshot` — same wording as before the refactor. Generically accurate. If you ever want screen-reader users to hear something more specific (e.g. "Swipe Watch iPhone demo video"), the component is the right place to take a new optional `alt` prop. Not for this PR.
2. The component has no TypeScript tests, but neither does any other Astro component in `src/components/`. Consistent with the repo.

PR #243, opened 2026-04-18
fix(mux): hide player chrome for GIF-alike hero (#237 follow-up)

nathanjohnpayne:
Follow-up to #242: Safari now autoplays (confirmed by user), but the player chrome (control bar + center play button) doesn't match the intended GIF-like hero.

Sets Mux's documented CSS vars `--controls: none` + `--center-controls: none` on the mux-player inline style, plus the `nohotkeys` attribute. Net effect: silent autoplay-loop video with no UI chrome, matching the feel the static GIF had before Phase 1.

Human authorized skipping the full review loop on this PR — fast-track merge + deploy to unblock Safari parity.

Authoring-Agent: claude

## Self-Review

Two changes: inline `style` gets `--controls: none; --center-controls: none;`, plus the `nohotkeys` boolean attr. No logic change, no new deps. Chromium path unaffected. Scope is one file, 1 insertion / 1 modified line.

PR #244, opened 2026-04-18
fix(mux): switch to <mux-video> for chromeless hero (Safari autoplay)

nathanjohnpayne:
Follow-up to #243. Hiding controls on `<mux-player>` via `--controls: none` regressed Safari autoplay (user confirmed on normal + private Safari after #243 deployed). Switching to `<mux-video>` — Mux's bare HTML5-like video primitive with HLS + Mux Data wiring but zero UI chrome. No controls to hide → no conflict with autoplay.

All attributes preserved (`autoplay="muted"`, `muted`, `loop`, `playsinline`, `preload="auto"`, `env-key`, `metadata-video-title`, `metadata-video-id`). Chromium preview verified: video upgrades, plays, loops, muted, no UI chrome.

Authoring-Agent: claude

## Self-Review

Scope: `ProjectMuxPlayer.astro` swaps the `<mux-player>` tag for `<mux-video>`, drops the now-irrelevant `--media-accent-color` / `--controls` / `nohotkeys` attributes, replaces the `` pattern with the standard HTML `poster` attribute (plus a bare `` child for no-JS fallback via unknown-custom-element light-DOM rendering), and updates the runtime import gate. `@mux/mux-video` promoted from transitive to direct dep. Not removing `@mux/mux-player` — deferred in case a future project wants the full UI.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "vitest": "^4.1.4"
   },
   "dependencies": {
-    "@mux/mux-player": "^3.11.8"
+    "@mux/mux-player": "^3.11.8",
+    "@mux/mux-video": "^0.30.6"
   }
 }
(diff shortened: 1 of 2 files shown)


nathanpayne-claude (approved):
Fast-track approval per human override on the #237 iteration loop. <mux-video> is the right primitive for a chromeless hero.

PR #247, opened 2026-04-18
fix(mux): use <mux-background-video> for Safari autoplay hero (#237)

nathanjohnpayne:
Per Mux's documented recipe for hero / GIF-replacement videos, `<mux-background-video>` is the purpose-built component. It handles the Safari HLS-manifest-load-vs-autoplay-policy timing race internally — the same race that neither `<mux-player>` (with chrome hidden) nor `<mux-video>` (with autoplay="muted") could reliably win.

Attribute surface swap:
- Drop `playback-id` + `autoplay="muted"` + `muted` + `loop` + `playsinline` + `preload` + `env-key` + `metadata-*`. mux-background-video handles these internally and doesn't take most as attributes.
- Add `src="https://stream.mux.com/{id}.m3u8"` (the form it expects).
- Add `max-resolution="720p"` to bound bitrate.
- Keep the static Mux JPEG thumbnail as the slotted poster child.
- Keep the `<noscript>` GIF fallback.

Mux Data analytics (Phase 2) is dropped in this PR — `<mux-background-video>` uses the separate `mux-embed` script, not an env-key attribute. Re-wiring Mux Data is a follow-up; autoplaying on Safari is higher priority.

Chromium preview verified: `video.paused=false`, `currentTime=14.05s`, `readyState=4`.

Refs #237.

Authoring-Agent: claude

## Self-Review

Two files touched: package.json (add @mux/mux-background-video@^0.2.3) and ProjectMuxPlayer.astro (full rewrite of the rendered tag + component doc). No CSS changes — .project-screenshot__mux class preserved.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "vitest": "^4.1.4"
   },
   "dependencies": {
+    "@mux/mux-background-video": "^0.2.3",
     "@mux/mux-player": "^3.11.8",
     "@mux/mux-video": "^0.30.6"
   }
--- a/src/components/ProjectMuxPlayer.astro
+++ b/src/components/ProjectMuxPlayer.astro
@@ -1,13 +1,21 @@
 ---
-// Hero video slot for project pages. When `playbackId` is set, renders a
-// chromeless <mux-video> element — Mux's bare video primitive with HLS +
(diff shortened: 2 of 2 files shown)


nathanpayne-claude (approved):
Fast-track per human override on #237. <mux-background-video> is Mux's documented recipe for this use case.

PR #246, opened 2026-04-18
fix(mux): static Mux JPEG poster disambiguates autoplay in Safari

nathanjohnpayne:
Swaps the animated GIF poster for Mux's static first-frame thumbnail (`image.mux.com/{id}/thumbnail.jpg?width=640&time=0`). Safari animates animated-GIF posters on `<video>`, making autoplay-succeeded visually indistinguishable from autoplay-blocked-but-poster-animating. A static JPEG removes the ambiguity: motion = video playing, frozen = autoplay blocked.

Chromium baseline preserved (video.currentTime advances, readyState 4).

Refs #237.

Authoring-Agent: claude

## Self-Review

One file, one attribute swap. Derives the JPEG URL from playbackId; no new dep, no new script. Keeps the GIF around for the noscript fallback.

PR #242, opened 2026-04-18
fix(mux): bare autoplay + preload=auto for Safari autoplay

nathanjohnpayne:
## Summary

Attempt 1 at [#237](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/237) — Safari doesn't autoplay the Mux player on `/projects/swipe-watch/` while Chrome does.

Two small changes on `<mux-player>` in [src/components/ProjectMuxPlayer.astro](src/components/ProjectMuxPlayer.astro):

1. **`autoplay="muted"` → `autoplay` (bare boolean).** `autoplay="muted"` is Mux-specific shorthand for "try autoplay, fall back to muted autoplay if denied." The HTML standard is a boolean `autoplay` attribute, and Safari's autoplay-policy gate reads the attribute via the standard path. `muted` is already set separately so playback stays silent. This matches [Mux's documented example](https://www.mux.com/docs/guides/autoplay-your-videos).
2. **Added `preload="auto"`.** Safari in some versions only attempts autoplay after metadata is available, and the default `preload="metadata"` can defer the first HLS segment. Explicit `preload="auto"` nudges Safari to fetch eagerly, closing a small timing window where `play()` fires before the player has anything to play.

## Verification

- Chromium baseline preserved. Built HTML emits `autoplay="true" preload="auto"` on `<mux-player>`; dev preview confirms video plays (`paused: false`, `currentTime` advances), muted, loops.
- Safari verification is on you — live deploy needed.

## Self-Review

**Correctness.** Two low-risk attribute changes, well-documented in the HTML / Mux spec references. The bare-boolean form `autoplay` JSX-renders to `autoplay="true"` in Astro (boolean attrs on custom elements are stringified), which is still treated as truthy by browsers per HTML spec.

**Regression risk.** Low.

- The `muted` attribute was already present, so removing `autoplay="muted"` doesn't change "is the video muted on autoplay" behavior.
- `preload="auto"` only accelerates fetch; it doesn't change what the player does post-fetch.
- Chromium verified unaffected.

**If this doesn't fix Safari.** Next attempts ranked:

1. Drop the animated GIF from ``. Safari has documented issues deferring `play()` when the poster is an ever-changing animated source. Swap for a static PNG extracted from the GIF's first frame.
2. Add a client-side `play()` fallback on the `canplay` event to force-start playback once the player has bufferable content.
3. Remove the `` entirely and let Mux's default poster (the Mux Player's auto-generated thumbnail) render instead.

[...]

PR #248, opened 2026-04-19
[codex] fix(mux): improve Swipe Watch background video quality

### LINKED ISSUE #237: bug(safari): mux-player does not autoplay on Swipe Watch in Safari

nathanjohnpayne:
## Summary

- restores the Swipe Watch hero to `mux-background-video`, the path that actually autoplayed in live Safari testing
- removes the previous `max-resolution="720p"` cap so Mux can select sharper portrait renditions for the narrow phone mockup
- upgrades the poster frame from a 640px still to a 1280px Mux thumbnail so the pre-upgrade / blocked state stays crisp on Retina screens
- removes the now-obsolete `mux-player` overlay-hiding path and updates the regression test to assert the background-video configuration instead

## Root Cause

The autoplay issue and the quality issue turned out to be separate problems.

Autoplay:
- live Safari testing showed the `mux-player` path still did not autoplay reliably enough for this page, even though Chromium and Playwright WebKit looked healthy
- `mux-background-video` remains the only implementation in this repo that has actually produced the desired GIF-like autoplay behavior in Safari for Swipe Watch

Quality:
- the working `mux-background-video` implementation in #247 explicitly limited the stream with `max-resolution="720p"`
- this asset's HLS manifest exposes much sharper portrait renditions than 720p, including 1178x2560 and 1206x2622 variants
- on a narrow phone mockup viewed on Retina/iPhone-class displays, that cap was the most likely reason the autoplaying video looked soft even though the playback path itself was correct

This PR keeps the Safari-safe playback path and removes the quality bottleneck instead of continuing to iterate on the failing `mux-player` path.

### #242
- Changed `ProjectMuxPlayer` to use `mux-player` with `autoplay="muted"` and `preload="auto"`
- Result: seemed promising in non-Safari validation, but the user's live Safari preview still did not autoplay reliably enough
- Why insufficient: did not hold up in the browser that mattered most for this bug

### #243
- Added `nohotkeys` plus broad player-chrome suppression (`--controls: none; --center-controls: none;`)
- Result: autoplay regressed in Safari
- Why insufficient: coupled UI suppression to playback behavior and removed the recovery path

### #244
- Replaced `mux-player` with `mux-video` to pursue a chromeless hero primitive
- Result: autoplay still unreliable and a stacked double-render appeared
- Why insufficient: moved to a lower-level path without solving Safari autoplay and introduced a new rendering bug

### #245
- Moved fallback markup out of `mux-video` into `\<noscript\>` to stop the stacked render
- Result: duplicate rendering improved, but autoplay still did not reliably start
- Why insufficient: fixed a secondary fallback/rendering bug, not the main autoplay requirement

### #246
- Replaced the animated GIF poster with a static Mux JPEG poster to make autoplay success observable
- Result: clarified the autoplay failure on the `mux-video` path
- Why insufficient: improved diagnosis only; did not fix playback behavior

### #247
- Switched to `mux-background-video`, which restored autoplay in Safari
- Result: autoplay worked, but the video looked softer than desired
- Why insufficient: solved playback reliability, but the quality tuning (`max-resolution="720p"`) was too conservative for this asset and layout

### Shared assumption across the failed fixes

The shared assumption was that the next step had to keep pushing toward `mux-player` or another chromeless primitive.

The more accurate framing after live Safari verification is:
- `mux-background-video` is the correct playback surface for this page
- the remaining problem is stream/poster quality, not missing UI suppression
- the fix is to keep the working Safari path and remove the artificial quality cap

## Validation

Automated:
- `npm test`
- `npm run test:e2e`

Targeted technical checks:
- confirmed the live preview HTML now renders `mux-background-video` again, not `mux-player`
- confirmed the `max-resolution` attribute is gone from the rendered hero markup
- confirmed the poster now uses `https://image.mux.com/.../thumbnail.jpg?width=1280&time=0`
- inspected the Mux HLS master playlist for this playback ID and verified it exposes portrait renditions above 720p

User-facing preview:
- live Safari testing reported that the background-video path autoplayed, while the `mux-player` path did not
- local preview was refreshed onto the background-video branch state for another Safari check before merge/deploy

Fixes #237.

Authoring-Agent: codex

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,6 @@
     "vitest": "^4.1.4"
   },
   "dependencies": {
-    "@mux/mux-background-video": "^0.2.3",
     "@mux/mux-player": "^3.11.8",
     "@mux/mux-video": "^0.30.6"
   }
--- a/package.json
+++ b/package.json
@@ -21,7 +21,6 @@
     "vitest": "^4.1.4"
   },
   "dependencies": {
-    "@mux/mux-player": "^3.11.8",
-    "@mux/mux-video": "^0.30.6"
+    "@mux/mux-background-video": "^0.2.3"
   }
 }
--- a/src/components/ProjectMuxPlayer.astro
+++ b/src/components/ProjectMuxPlayer.astro
@@ -1,21 +1,12 @@
 ---
 // Hero video slot for project pages. When `playbackId` is set, renders
-// Mux's dedicated background-video component — the one Mux built for
-// this exact use case (hero / GIF-replacement / autoplay-muted-loop).
+// <mux-player> with the conservative Safari-friendly autoplay config that
+// already worked in the #237 audit. The player UI stays available for
+// recovery, but CSS hides only the centered play overlay so the hero no
+// longer looks paused on top of an autoplaying clip.
 //
-// Why <mux-background-video> and not <mux-video> or <mux-player>?
-// - <mux-player> is a full player with controls; hiding the chrome
-//   via CSS vars regressed Safari autoplay.
-// - <mux-video> is a low-level HTML5-like primitive; on Safari the
-//   native HLS manifest load races against Safari's autoplay policy
-//   check, silently leaving the video paused on the poster.
-// - <mux-background-video> is purpose-built for hero videos: no UI
-//   chrome, internal autoplay orchestration that handles the HLS-on-
-//   Safari timing race, and the documented component for this
-//   use case in Mux's own docs.
-//
-// When `playbackId` is unset, falls back to the `fallbackSrc` image
-// so the hero never goes blank.
+// When `playbackId` is unset, falls back to the `fallbackSrc` image so the
+// hero never goes blank.
 
 interface Props {
   playbackId?: string;
@@ -26,43 +17,36 @@ interface Props {
 
 const { playbackId, fallbackSrc, title, slug } = Astro.props;
 
+// Mux Data env key. Baked into the build via `import.meta.env`; when unset,
+// Astro omits the attribute and the player emits no analytics.
+const muxEnvKey = import.meta.env.PUBLIC_MUX_ENV_KEY;
+
 const altText = `${title} product screenshot`;
-
-const streamSrc = playbackId
-  ? `https://stream.mux.com/${playbackId}.m3u8`
-  : undefined;
-
-// Static first-frame poster from Mux's image API. JPEG at 640px is
-// sharper than the GIF rendered at 320px and makes the autoplay-vs-
-// not-autoplaying signal unambiguous (motion = playing; frozen =
-// blocked).
-const posterSrc = playbackId
-  ? `https://image.mux.com/${playbackId}/thumbnail.jpg?width=640&time=0`
-  : undefined;
 ---
 
 {playbackId ? (
-  <>
-    <mux-background-video
-      src={streamSrc}
-      max-resolution="720p"
-      class="project-screenshot__mux"
-    >
-      <img src={posterSrc} alt={altText} />
-    </mux-background-video>
-    <noscript>
-      <img src={fallbackSrc} alt={altText} loading="eager" />
-    </noscript>
-  </>
+  <mux-player
+    playback-id={playbackId}
+    autoplay="muted"
+    muted
+    loop
+    playsinline
+    preload="auto"
+    env-key={muxEnvKey}
+    metadata-video-title={title}
+    metadata-video-id={slug}
+    class="project-screenshot__mux"
+    style="--media-accent-color: var(--project-accent); --media-primary-color: var(--project-accent);"
+  >
+    <img slot="poster" src={fallbackSrc} alt={altText} />
+  </mux-player>
 ) : (
   <img src={fallbackSrc} alt={altText} loading="eager" />
 )}
 
 <script>
-  // Register <mux-background-video> on pages that use it. Dropping
-  // mux-player and mux-video from the runtime on this page — this
-  // component is the sole consumer.
-  if (document.querySelector('mux-background-video')) {
-    import('@mux/mux-background-video/html');
+  // Register <mux-player> only on pages that actually contain one.
+  if (document.querySelector('mux-player')) {
+    import('@mux/mux-player');
   }
 </script>
--- a/src/components/ProjectMuxPlayer.astro
+++ b/src/components/ProjectMuxPlayer.astro
@@ -1,9 +1,17 @@
 ---
 // Hero video slot for project pages. When `playbackId` is set, renders
-// <mux-player> with the conservative Safari-friendly autoplay config that
-// already worked in the #237 audit. The player UI stays available for
-// recovery, but CSS hides only the centered play overlay so the hero no
-// longer looks paused on top of an autoplaying clip.
+// Mux's dedicated background-video component — the one Mux built for
+// this exact use case (hero / GIF-replacement / autoplay-muted-loop).
+//
+// Why <mux-background-video> and not <mux-video> or <mux-player>?
+// - <mux-player> is a full player with controls; hiding the chrome
+//   regressed Safari autoplay in real-device testing.
+// - <mux-video> is a low-level HTML5-like primitive; on Safari the
+//   native HLS manifest load races against Safari's autoplay policy
+//   check, silently leaving the video paused on the poster.
+// - <mux-background-video> is purpose-built for hero videos: no UI
+//   chrome, internal autoplay orchestration, and the Safari behavior
+//   this project actually needs.
 //
 // When `playbackId` is unset, falls back to the `fallbackSrc` image so the
 // hero never goes blank.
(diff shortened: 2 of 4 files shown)


nathanjohnpayne:
@coderabbitai review

nathanjohnpayne:
**Agent disagreement resolved.** Reviewers are no longer in a mixed state. Removing `needs-human-review` label.

nathanjohnpayne:
@coderabbitai review

nathanpayne-codex (approved):
Internal review as nathanpayne-codex on commit fd265a7.

Checked the final diff with focus on three risk areas: preserving Safari-safe autoplay attributes on `mux-player`, hiding only the centered overlay while keeping bottom recovery controls, and omitting `env-key` when `PUBLIC_MUX_ENV_KEY` is blank. I also confirmed the follow-up fix after CodeRabbit and re-ran `npm test` plus `npm run test:e2e`. No additional findings from this reviewer pass.

nathanjohnpayne:
## External Review Required

**PR:** #248 — [codex] fix(mux): hide Swipe Watch center play overlay
**Branch:** codex/swipe-watch-hide-center-play-overlay
**Author Agent:** codex

### Summary
This PR moves the Swipe Watch hero back to `<mux-player>`, which was the Safari-autoplay-capable path from the earlier audit, and hides only the centered play overlay instead of stripping the full player UI. The bottom controls remain available as a recovery path if Safari blocks autoplay due to browser or user policy. A follow-up commit also normalizes a blank `PUBLIC_MUX_ENV_KEY` to omit `env-key` entirely, matching the documented disabled state.

### Focus Areas
- Verify Safari/WebKit still treats the `mux-player` config as autoplay-safe with `autoplay="muted"`, `muted`, `loop`, `playsinline`, and `preload="auto"`.
- Verify the CSS change removes only the centered overlay and does not suppress the bottom controls or introduce regressions in hover/focus/player accessibility states.
- Verify the `env-key` binding and poster-slot fallback behave correctly when `PUBLIC_MUX_ENV_KEY` is blank or unset.

### Observations from Internal Review
- CodeRabbit flagged one substantive issue: Astro would render `env-key=""` when `PUBLIC_MUX_ENV_KEY=`. Fixed in follow-up commit `fd265a7` by normalizing blank values to `undefined`.
- Internal review as `nathanpayne-codex` found no additional issues after re-running `npm test` and `npm run test:e2e`.
- This is a browser-sensitive fix on a bug with several failed prior attempts, so even though the code change is relatively narrow, the review should scrutinize subtle autoplay/control regressions carefully.

### Suggested External Reviewer
nathanpayne-claude

### Rationale for External Review
User-requested heightened scrutiny on a historically flaky Safari autoplay regression. This handoff is precautionary/manual rather than triggered by the default line-count or protected-path threshold.

nathanpayne-claude (approved):
Reviewed as `nathanpayne-claude` per the external-review handoff at [#248 (comment)](https://github.com/nathanjohnpayne/nathanpaynedotcom/pull/248#issuecomment-4274844464).

**Approach.** Checked out the branch locally, ran the dev server, probed mux-player's shadow DOM directly to verify the `::part()` selector targets what the PR description claims it targets. The CSS `::part()` selector is the load-bearing change here — easy to get wrong since Media Chrome's part names aren't in the Mux docs — so observation beats reasoning for this one.

**`::part(center play button)` targeting is correct.** Mux player exposes the centered play-button overlay as `<media-play-button part="center play button">` in its shadow DOM. The selector matches exactly one element, computed `display: none`, rect `0 × 0`, `visible: false`. No other element in the tree has those three tokens in its `part` attribute — the bottom play button is `part="bottom play button"` (different tokens), the center chrome container is `part="layer centered-layer center centered chrome"` (different shape), so there's no accidental collision.

**Bottom controls preserved.** The `control-bar bottom` element is still in the shadow DOM with `display: flex` and only dims to `opacity: 0` on idle — Mux's standard auto-hide-on-idle / show-on-hover UX. The PR body claims "Bottom controls remain for manual recovery"; verified: they're in the tree, dormant, summonable on hover. Not removed.

**Autoplay config unchanged from the known-good #242 state.**

| attr | value |
|---|---|
| `autoplay` | `muted` |
| `muted` | present |
| `loop` | present |
| `playsinline` | present |
| `preload` | `auto` |
| `env-key` | `6e3df7ilf13pn04stchr551h3` |

Video upgrades cleanly, `video.paused === false`, `video.currentTime` advances to 51s+ during probing, `readyState: 4` (HAVE_ENOUGH_DATA), `muted: true`, `loop: true`. No `--controls: none` overrides in the inline style — only the two accent CSS vars, matching the config that was Safari-autoplay-verified before [#243](https://github.com/nathanjohnpayne/nathanpaynedotcom/pull/243) regressed it. Keeps us out of the trap where hiding all controls destabilized autoplay.

**CodeRabbit's Major finding is correctly addressed.** The finding: `env-key=""` would be rendered when `.env.local` contains `PUBLIC_MUX_ENV_KEY=` (the documented disabled state). Astro renders empty strings as `attr=""`, only `undefined`/`null` omit the attribute. The fix in `fd265a7` is:

```ts
const muxEnvKey = import.meta.env.PUBLIC_MUX_ENV_KEY?.trim() || undefined;
```

This is literally CodeRabbit's suggested diff. Handles:
- `undefined` → `undefined` (attr omitted)
- `""` → `""` → `undefined` via `||` (attr omitted)
- `"   "` → `""` after trim → `undefined` (attr omitted)
- `"6e3df..."` → `"6e3df..."` (attr rendered)

Cleanly closed.

**Regression test coverage.** The new `tests/project-pages.test.js` case asserts the mux-player attrs + poster-slot wiring + explicit `mux-background-video` non-presence — nice regression guard against re-landing [#247](https://github.com/nathanjohnpayne/nathanpaynedotcom/pull/247)-style component swaps. Doesn't cover the `::part()` CSS (jsdom doesn't parse `::part` selectors), but that's an acceptable gap — the CSS rule is short and visually observable.

**Dependency hygiene.** `package.json` drops `@mux/mux-background-video` now that the component isn't used. Keeps `@mux/mux-video` even though nothing in the component tree references it — left as a transitive/reserved dep. Minor: could drop that too, but it's not worth blocking on and it's a transitive dep of `@mux/mux-player` anyway.

**Non-blocking observations.**

1. The PR description and the inline CSS comment both frame this as Safari-verified; the verification is actually on Chromium in the dev preview here, and Safari-specific re-test is still pending the user. That's fine — the autoplay config is the one that worked in the earlier [#242](https://github.com/nathanjohnpayne/nathanpaynedotcom/pull/242) Safari audit, and the CSS-only center-overlay hide shouldn't interact with autoplay the way `--controls: none` did in [#243](https://github.com/nathanjohnpayne/nathanpaynedotcom/pull/243).
2. The open items in [#237](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/237) (layout regression on the Background Video path, Mux Data re-wire) are implicitly resolved by this PR since it reverts to the `<mux-player>` path that never had those issues. Worth noting in the issue closeout comment if this lands.

nathanjohnpayne:
**Agent disagreement resolved.** Reviewers are no longer in a mixed state. Removing `needs-human-review` label.

nathanjohnpayne:
@coderabbitai review

nathanpayne-codex (approved):
Internal review as nathanpayne-codex on commit 43d2eb9.

Reviewed the pivot back to `mux-background-video` with focus on three areas: preserving the Safari-working autoplay surface, removing the artificial 720p quality cap without changing the layout contract, and keeping the regression test aligned with the rendered hero markup. CodeRabbit returned no actionable findings on this head, and the branch re-passed `npm test` plus `npm run test:e2e`. No additional findings from this reviewer pass.

nathanjohnpayne:
**Agent disagreement resolved.** Reviewers are no longer in a mixed state. Removing `needs-human-review` label.

Issue #305, opened 2026-04-28
Homepage: clockwise pulse sequence to signal block interactivity on load


nathanjohnpayne:
## Related

Supersedes #271 (homepage motion nudges). The earlier ticket proposed two
inactivity-triggered nudges as polite hints; this ticket reframes the
problem as critical discoverability and replaces the inactivity pattern
with a single on-load sequence across all four interactive blocks.

## Problem

The homepage loads as a pure Mondrian composition: four labeled colored blocks (red/Nathan Payne, yellow/Vibe Coding, dark gray/Community, blue/Connect) plus filler blocks. All four primary blocks are interactive—mousing over any of them reveals the content inside—but at rest the page reads as art, not as a navigable site.

A first-time visitor with no signal that the blocks are interactive may look at the page and leave. The discoverability problem isn't "this one block needs a hint"—it's "this entire page needs to communicate that anything is clickable at all."

This ticket adds a single on-load animated sequence that pulses each interactive block in turn, telling the visitor "these are tiles, click them." After the sequence plays once, the page is quiet.

## Fix

On page load, after a brief settle delay, the four primary interactive blocks pulse in clockwise sequence. Each pulse is subtle—a brightness/saturation shift on the block's fill—and the sequence as a whole takes roughly 1.4 seconds.

### Sequence

Clockwise from the upper-left, four blocks total:

1. **Nathan Payne (red, upper-left)** — pulses first
2. **Vibe Coding (yellow, upper-right)** — pulses second
3. **Connect (blue, lower-right)** — pulses third
4. **Community (dark gray, lower-left)** — pulses fourth

Filler blocks (white, light gray) do not pulse. They aren't interactive and including them would dilute the signal.

The clockwise order traces the page's perimeter, reinforcing the Mondrian-frame quality. The user's eye moves around the page once, and now they know the entire frame is alive.

### Pulse mechanic

Each pulse is a brief brightness and saturation shift on the block's colored fill, then back to rest. Specifically:

- **Filter:** `filter: brightness(1.08) saturate(1.05)` at peak, returning to `filter: none` at rest
- **Duration per pulse:** 250ms (125ms in, 125ms out)
- **Easing:** ease-out
- **Inter-pulse gap:** 120ms (next block begins 370ms after previous block began, allowing slight overlap on the trailing edge)

Total sequence duration: roughly 1.4 seconds (4 pulses × 250ms + 3 gaps × 120ms ≈ 1360ms).

[...]

PR #307, opened 2026-04-28
fix(home): pulse sequence collapses to 1ms in production builds

nathanjohnpayne:
User reported the on-load clockwise pulse sequence (#305) was not visibly firing on the deployed site after PR #306 shipped.

## Root cause
Vite/lightningcss minifies CSS durations to whichever serialization is shorter, so the new tokens emit as `.7s` / `.37s` / `.25s` in production (verified via the live `global.CRTD5GZ-.css`). The JS parser in `src/pages/index.astro` used a bare `parseFloat` and read `.7s` as `0.7` — interpreted as milliseconds — so:

- `PULSE_INITIAL_DELAY = 0.7` instead of `700`
- `PULSE_INTERVAL = 0.37` instead of `370`
- `PULSE_DURATION = 0.25` instead of `250`

All four pulses fired within ~1ms of each other on load and the `.panel--pulsing` class was removed before the keyframe could play — hence "no visible animation". Dev mode masked the bug because Vite does not minify durations there.

## Fix
Make `readMs` unit-aware. Honors both `ms` and `s` and falls back gracefully on empty/garbage input. Verified in node against the production-minified values:

```
"700ms"  => 700        ".7s"   => 700
"370ms"  => 370        ".37s"  => 370
"250ms"  => 250        ".25s"  => 250
""/null/"abc"/"700"   => fallback
```

Authoring-Agent: claude

## Self-Review
- Correctness: regex `^([+-]?[\d.]+)\s*(ms|s)\s*$` matches both serializations and rejects malformed input cleanly. Fallback path preserved.
- Regression risk: `readMs` is only called by the on-load pulse sequence; no other callers.
- Test coverage: validated against the actual production CSS values + edge cases via a node REPL run captured in the commit body.
- Security/dependency hygiene: none.

## Test plan
- [ ] Hard reload nathanpayne.com on a deploy of this commit. Red → yellow → blue → black pulse clockwise over ~1.4s starting 700ms after the page becomes visible.
- [ ] Reduced motion still no-ops.
- [ ] Dev mode (`npm run dev`) still works (was always working).

### DIFF
--- a/src/pages/index.astro
+++ b/src/pages/index.astro
@@ -463,11 +463,18 @@ const homepageJsonLd = {
       // Read spec-locked timings from CSS custom properties so the
       // values stay controllable from one place (src/styles/global.css
       // :root) and the JS doesn't drift from the CSS keyframe duration.
+      // Vite/lightningcss minifies durations to whichever serialization
+      // is shorter — `700ms` is emitted as `.7s` in production — so the
+      // parser has to honor both units. parseFloat alone would read
+      // `.7s` as 0.7 and collapse the entire 1.4s sequence to 1.4ms.
       const motionTokens = getComputedStyle(document.documentElement);
       const readMs = (name, fallback) => {
         const raw = motionTokens.getPropertyValue(name).trim();
-        const n = parseFloat(raw);
-        return Number.isFinite(n) ? n : fallback;
+        const m = raw.match(/^([+-]?[\d.]+)\s*(ms|s)\s*$/i);
+        if (!m) return fallback;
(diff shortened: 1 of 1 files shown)


nathanpayne-claude (approved):
LGTM. Real bug, clean diagnosis (Vite/lightningcss serializes 700ms as .7s which parseFloat misreads as 0.7 ms). The unit-aware regex handles both forms and falls back on garbage. Caller surface is one helper used in one place. Dev mode masking the bug explains why it shipped clean to staging-equivalent.

PR #308, opened 2026-04-28
fix(home): pulse waits for label fade-in to complete before firing

nathanjohnpayne:
User noted a residual flash on load: panel labels were still mid-fade-in (the font-readiness gate from #306) when the on-load pulse sequence (#305) started its first brightness shift. The two animations overlapped and read as a single hard flash.

## Fix
Chain the pulse on `document.fonts.ready` so the labels fully resolve in Cormorant Garamond before any motion starts. Failsafe path (`setTimeout` 1700ms) ensures the cue still fires if `document.fonts` is unavailable or never resolves — matches `BaseLayout`s 1500ms font gate timeout plus a 200ms safety margin.

## Token retune
`--pulse-initial-delay`: 700ms → 300ms.
- Was "settle from page load"; now "breath after panel labels finish fading in before the sequence fires".
- Labels themselves fade over `--motion-hover` (~170ms) once `fonts.ready` resolves; 300ms leaves the page visibly settled before motion.
- With cached fonts, total time-to-first-pulse is ~480ms (was 700ms). With cold fonts it is bounded by `fonts.ready + 300ms` (typically <800ms).

Issue #305 originally specified 700ms as a settle delay from load on the assumption that nothing else was animating during that window. With the font-gate fade-in now occupying the early window, the spec intent ("page demonstrates its interactivity once it has settled") is better met by chaining on the actual settle event rather than guessing the duration up front.

Authoring-Agent: claude

## Self-Review
- **Correctness:** `runPulseSequence` now wraps the existing `setTimeout` chain in a `start` closure that resolves on either `document.fonts.ready` or the 1700ms failsafe. Inner `scheduled` flag prevents double-fire if both paths resolve. The outer `pulseFired` guard still keeps it once-per-load.
- **Regression risk:** Reduced-motion path unchanged (still no-ops). Background-tab visibility deferral unchanged. Failsafe matches the upstream font gate so the worst-case path still completes.
- **Test coverage:** None for these classes; verified the production CSS now serializes `--pulse-initial-delay` as `.3s` and the `readMs` parser (PR #307) reads it as 300ms.
- **Security / dependency hygiene:** None.

## Test plan
- [ ] Hard reload nathanpayne.com — labels fade in in Cormorant Garamond, then ~300ms later the four panels pulse clockwise (red → yellow → blue → black). No overlap between the two motions.
- [ ] Reduce Motion enabled — pulse no-ops; label fade transition collapses to instant per existing rule.
- [ ] Load in a background tab, then surface — sequence still defers to `visibilitychange`.

### DIFF
--- a/src/pages/index.astro
+++ b/src/pages/index.astro
@@ -476,9 +476,13 @@ const homepageJsonLd = {
         if (!Number.isFinite(n)) return fallback;
         return m[2].toLowerCase() === 's' ? n * 1000 : n;
       };
-      const PULSE_INITIAL_DELAY = readMs('--pulse-initial-delay', 700);
+      const PULSE_INITIAL_DELAY = readMs('--pulse-initial-delay', 300);
       const PULSE_INTERVAL = readMs('--pulse-interval', 370);
       const PULSE_DURATION = readMs('--pulse-duration', 250);
+      // Outer cap: BaseLayout's font gate falls open after 1500ms even if
+      // document.fonts.ready never resolves; pulse should still fire in
+      // that case, just from the failsafe path.
+      const FONTS_READY_FAILSAFE = 1700;
       let pulseFired = false;
 
       function pulsePanel(className) {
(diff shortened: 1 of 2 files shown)


nathanpayne-claude (approved):
LGTM. Right diagnosis (label fade + first pulse overlapping reads as a flash) and right fix (chain pulse on document.fonts.ready). The 1700ms failsafe matches the upstream BaseLayout font gate's 1500ms timeout + a 200ms margin, so a stalled fonts.ready can't strand the discoverability cue. Inner scheduled flag is the right belt-and-suspenders for the dual-resolution path. Token retune from 700ms to 300ms is a semantic shift documented in the comment and PR body — same-or-shorter time-to-first-pulse in both cached and cold paths.

PR #325, opened 2026-05-09
fix(home): keep Connect tile visible when Community is hovered

nathanjohnpayne:
Authoring-Agent: claude

Fixes a regression flagged by the user via screenshot: hovering the Community panel on the homepage hides the Connect tile entirely, violating the "all four colored Mondrian tiles always visible" invariant the composition is built around.

## What was happening
The community-focus grid template collapsed track 8 to 0:

```css
.mondrian[data-focus="community"] {
  grid-template-rows:
    …
    var(--cell-h-community, 401px)    /* track 6 — community cell */
    0                                 /* track 7 — line absorbed */
    0                                 /* track 8 — absorbed (was the bug) */
    var(--line);                      /* track 9 */
}
```

Connect's panel has `grid-row: 8` (cols 6–8) by default. With track 8 = 0, Connect rendered at height 0 — invisible to users.

The other multi-row spanning panels (About in about-focus, Builds in projects-focus) work because their absorbed tracks (3 and 4) sit inside their own column footprint with no other panel placed on those tracks. Community is the asymmetric case: track 8 is shared between Community's tail (col 2) and Connect (cols 6–8), so blanket-absorbing it hides Connect.

## The fix (16 lines, [src/styles/global.css:217-219, 222-235](src/styles/global.css))
Splits the absorption asymmetrically:

1. **Track 7 stays absorbed** (`0`). The helper line lives inside Community on its own column, matching the about-focus and projects-focus pattern.
2. **Track 8 restored to `minmax(3.4rem, 0.95fr)`** — same value about-focus uses for Connect's row. Connect renders at its non-focused height.
3. **New override `.mondrian[data-focus="community"] .panel--black { grid-row: 6 / 8 }`** stops Community from spanning into track 8 in col 2, so its column doesn't bleed into Connect's row. Mirrors the existing `.mondrian[data-focus="projects"] .panel--yellow { grid-row: 2 / 5 }` override pattern.

A multiline comment explains the asymmetry inline so future readers don't try to "restore symmetry" by re-collapsing the row.

## Testing — focus-state audit
Programmatically swept all five focus states (`none`, `about`, `projects`, `community`, `connect`) and measured each panel's bounding rect. Every panel is non-zero in every state after the fix. The community-focus row specifically:

| Panel | Before | After |
|---|---|---|
| About | 592 × 314 | 592 × 202 |
| Projects | 234 × 165 | 234 × 104 |
| Community | 486 × 512 | 486 × 512 |
| **Connect** | **234 × 0** | **234 × 112** |

About and Projects shrank slightly (the new minmax(0.95fr) at track 8 reduces the share available for the top minmax tracks), but both remain visible at sensible proportions. The other four focus states are unchanged — the rule is scoped to `[data-focus="community"]`.

- [x] Tests pass — CSS-only change; no logic / schema / route / build-script change. The existing Playwright responsive smoke tests should sail through (they verify panel layout under various viewports; this fix improves layout, doesn't regress it).
- [x] Build succeeds — verified locally; HMR picked up the rule cleanly.
- [x] Manual verification completed — programmatic audit + visual screenshot in the dev preview (community-hover now shows red About, yellow BUILDS, expanded Community panel, AND blue CONNECT all simultaneously).

## Self-Review
- [x] Correctness: bug reproduced before the fix (Connect height 0 in community-focus), absent after (Connect height 112). Audit confirms no other state regressed.
- [x] Regression risk: low. Rule is scoped exclusively to `[data-focus="community"]`. The override of `.panel--black` grid-row is also gated to community-focus; in all other states the panel uses its default grid-row. CSS selectors keying on `data-panel="community"` (test selectors, the JS focus-state machine) are untouched.
- [x] Style and conventions: matches the existing comment style (referenced PR numbers, design intent before mechanics) and the existing override-rule pattern (`projects-focus` → `.panel--yellow`). Inline comment explains the asymmetry so the absorption rule isn't "re-fixed" later.
- [x] Test coverage: no new code paths; existing Playwright suite covers homepage layout.
- [x] Security and dependency hygiene: zero dependency changes, no new scripts, no new external links.

## Review path
+16 / -1 in a single CSS file. Sub-threshold; no protected paths. Internal review only — merge as nathanjohnpayne after CodeRabbit clears.

## Out of scope
- The slightly thin Connect tile in projects-focus (~45px tall at this viewport size) is a separate proportional question — that state shows all four panels but Connect is unusually compressed. Not visible-vs-hidden, so not a blocker. Leaving for a separate proportional-tuning ticket if desired.

### DIFF
--- a/src/styles/global.css
+++ b/src/styles/global.css
@@ -214,8 +214,23 @@ html, body {
     var(--line)                       /* track 5 — single line above community */
     var(--cell-h-community, 401px)    /* track 6 — community cell */
     0                                 /* track 7 — line absorbed (inside community) */
-    0                                 /* track 8 — absorbed */
+    minmax(3.4rem, 0.95fr)            /* track 8 — connect's row (cannot be absorbed; connect lives here) */
     var(--line);                      /* track 9 — bottom boundary */
+}
+
+/* Community-focus is the asymmetric absorption case (#322 follow-up).
+   The other multi-row spanning panels (about, projects) absorb tracks
+   that no other panel occupies, so they collapse helper rows to 0
+   uniformly. Community spans tracks 6 / 7 / 8, but Connect lives on
+   track 8 — collapsing it would hide Connect entirely, breaking the
+   "all four colored tiles always visible" invariant. The fix:
+     1. Keep track 7 absorbed (the helper line stays inside community).
+     2. Restore track 8 to a real minmax so Connect renders.
+     3. Stop community from spanning into track 8 so it doesn't bleed
+        into Connect's row in col 2.
+   Mirrors the projects-focus override of `.panel--yellow` grid-row. */
+.mondrian[data-focus="community"] .panel--black {
+  grid-row: 6 / 8;
 }
 
 .mondrian[data-focus="connect"] {


nathanpayne-claude (commented):
Reviewer: nathanpayne-claude (self-review on a claude-authored PR — `--comment` per REVIEW_POLICY.md)

## Scope check
+16 / -1 in a single CSS file. Adds one new minmax track value, one new selector rule (4 lines + 11-line comment block). Below the 300-line external-review threshold; no protected paths. Internal review only.

## Correctness
The diagnosis is sound: pre-fix DOM measurement showed Connect at 234×0 in community-focus while every other panel was non-zero. The CSS template at line 217 explicitly set track 8 to 0; Connect's `grid-row: 8` lives there. The two are causally linked.

The fix has three coherent parts:
1. Track 8 restored to `minmax(3.4rem, 0.95fr)` — same value about-focus uses for the analogous row, so Connect's height in community-focus matches its height in about-focus (~112px in current viewport, 151px in about-focus). Reasonable.
2. Track 7 stays at 0 — preserves the design intent of absorbing Community's helper-line track into its body, mirroring how about-focus absorbs track 3.
3. `.panel--black { grid-row: 6 / 8 }` override — stops Community from spanning into track 8 in its column, which would have left an awkward empty band below Community in col 2 once track 8 became non-zero. Mirrors the existing `[data-focus="projects"] .panel--yellow { grid-row: 2 / 5 }` override.

## Audit results
Measured all five focus states post-fix:
- `none`: about 474×576, projects 352×427, community 317×250, connect 352×157 ✓
- `about`: about 598×571, projects 228×571, community 421×255, connect 228×151 ✓
- `projects`: about 195×792, projects 631×792, community 95×79, connect 631×45 ✓
- `community`: about 592×202, projects 234×104, community 486×512, connect 234×112 ✓ (was 234×0)
- `connect`: about 332×265, projects 494×137, community 178×561, connect 494×440 ✓

All four panels at non-zero dimensions in all five states. The two states where Connect appears compressed (`projects` at 45px, `community` at 112px) are clearly visible bands of color — not the previous "hidden" regression. The about/projects rows shrank slightly in community-focus (314→202, 165→104) because the row template now has more total fr to distribute; both remain readable at typical viewport sizes.

## Regression risk
- **CSS scope**: rule is gated on `[data-focus="community"]`. No other focus state's template was edited. Default `.mondrian` rule untouched. The `.panel--black` override is also gated, so non-community-focus states use the default `grid-row: 6 / 9`.
- **JS handlers**: the focus-state machine in `src/pages/index.astro` keys on `dataset.focus` reads/writes; nothing in the fix renames or moves attributes. No JS change.
- **OG generation, sitemap, integrations**: untouched.
- **Tests**: existing Playwright responsive tests cover homepage layout. They should pass — the change adds visibility, doesn't remove it. (And if a test previously asserted Connect's height as 0 in community-focus, it was asserting the bug; that test would correctly fail and need updating.)

## Style / conventions
- The 11-line comment block above the new override follows the existing convention (referenced issue, design intent ahead of mechanics, future-reader-friendly framing). Matches the comments at lines 132–161 and 1030–1042.
- Uses the same minmax syntax and `var(--line)` placeholders the rest of the file uses.
- Pulls the `0.95fr` constant from about-focus's track 8 deliberately — keeps Connect's height visually comparable across the two focus states where it's a non-focused panel.

## Test coverage
No new code paths. The fix would be defensively backed by a Playwright assertion: "in community-focus, all four `[data-panel]` elements have `getBoundingClientRect().height > 0`." Worth adding as a follow-up if the team wants explicit regression coverage; not in this PR's scope.

## Security / dependency hygiene
Zero. CSS-only.

## Verification
Visual confirmation in the dev preview: community-hover screenshot shows red About, yellow BUILDS, expanded Community panel, and blue CONNECT tile all simultaneously visible. Screenshot included in the PR description.

Cleared from my side. Awaiting CodeRabbit, then merge as nathanjohnpayne.

PR #340, opened 2026-05-10
fix(home): harmonize Community section body typography with Builds (#339)

### LINKED ISSUE #339: Homepage: harmonize Community section typography with Builds section
## Problem

When expanded, the Community section (Giving & Governing) renders body text at a noticeably smaller size than the Builds section. Both sections have the same content structure—intro paragraph, list of items with titles + descriptions, small caps footer—but body text in Community is smaller than the equivalent text in Builds.

Same content type, two different sizes. The site has been getting more typographically consistent over recent work (shared kicker treatment, shared tag rendering, shared metadata vocabulary); body text should follow the same principle.

## Fix

Match the Community section's body typography to the Builds section. Specifically:

- **Intro paragraph** — "Community organizing—identifying shared stakes..." should match the size, weight, and line-height of "Every project started as a real problem..." in Builds.
- **Item titles** — "Larkin Street Youth Services →" should match "Mergepath →" sizing in Builds.
- **Item amounts** ("$16,684 raised") — should be styled at body size, matching how Builds renders any equivalent item-level metadata.
- **Item descriptions** — "$16,684 raised for Safe & Seen—a campaign for LGBTQ+ youth..." should match the description size used in Builds for "A deterministic repository standard that keeps humans..."
- **Footer (SCOPE / STACK)** — already appears to match; verify no drift.

Other Community-specific styling (the section's overall expanded-state size, the right-rail Connect block visible alongside, etc.) is unchanged.

## Likely cause

The HTML structure for Community uses a `<ul>` of items, while Builds uses separate block elements per project. If the Community `<ul>` (or its `<li>` children) inherits from a list-typography rule that sets a smaller font size than the section's default body size, that would explain the discrepancy.

Implementation hint: check whether the Community list items inherit from a global `ul`/`li` rule, or whether the Community section template applies a smaller size locally. The fix is probably either:
- Override the list inheritance in the Community section so items render at section body size, or
- Restructure the Community items to use the same block elements as Builds projects (more invasive but more consistent at the structural level)

The first approach is cheaper. The second is cleaner long-term. Pick based on how much restructuring feels appropriate.

## Acceptance criteria

- [ ] Community intro paragraph renders at the same size, weight, and line-height as the Builds intro paragraph
- [ ] Community item titles render at the same size as Builds item titles
- [ ] Community item descriptions render at the same size as Builds item descriptions
- [ ] Community item amounts ("$X,XXX raised") render at body size, consistent with Builds
- [ ] SCOPE footer typography matches STACK footer typography
- [ ] No regressions in mobile breakpoint—both sections should still render legibly when stacked vertically
- [ ] No unintended downstream effects on other pages that may share list typography (verify project detail pages, blog posts, etc. didn't have a dependency on the rule being changed)

## Notes

The principle worth capturing somewhere in the design system, if not already documented: **same content type uses the same typography across the site, regardless of which container it's in.** Body text is body text. The right reasons to vary text size are role-based (heading vs. body vs. caption vs. metadata), not container-based.

Specifically, the right reason to size text down isn't "the container is more compressed" or "there's more content to fit." If a section's content runs long at the standard size, the fix is usually narrower columns, more aggressive editing, or a different layout—not smaller text.

If during implementation it turns out the Community size was deliberately chosen to fit more content in less vertical space, that's a layout problem, not a typography problem. Surface it rather than working around it. The Community section shouldn't have to fit a fixed height; its container should size to its content the same way Builds does.

## Out of scope

- Restructuring the Community section's HTML (use the lighter override approach unless there's a stronger reason)
- Changing Builds typography to match Community (Builds is the reference; Community moves to it)
- Restyling section headings ("Giving & Governing", "Builds")—those already match
- Restyling links (arrow, underline, hover)—handled by other tickets
- Changing the Community section's overall size, padding, or layout—only the body text typography is in scope

nathanjohnpayne:
Authoring-Agent: claude

Closes #339. Community section body text now matches Builds on size, line-height, weight, and opacity. The user-perceptible discrepancies (intro and item titles rendering larger; item descriptions rendering smaller AND faded at opacity 0.72) are gone.

## Empirical before/after at 1440×900 (Builds reference column on the left)

| Element | Builds | Community before | Community after |
|---|---|---|---|
| Intro paragraph | 13.6px / lh 20.13 / op 1 | 15.2px / lh 21.58 | **13.6px / lh 20.13 / op 1** ✓ |
| Item title | 13.6px / 700 / Cormorant Garamond | 15.04px / 700 / Cormorant Garamond | **13.6px / 700 / Cormorant Garamond** ✓ |
| Item description | 13.6px / lh 20.13 / op 1 | 13.44px / lh 18.55 / op 0.72 | **13.6px / lh 20.13 / op 1** ✓ |
| Item amount | n/a | 13.44px / 500 | 13.6px / 500 (now at body size, weight preserved as metadata cue) |

Mobile (375×812 stack) verified identical: every body element across both sections at 13.6px, opacity 1.

## Specific edits ([src/styles/global.css](src/styles/global.css))

- `.panel--black .content-inner > p` — `font-size: 0.85rem`; `line-height: 1.48`. Was 0.95rem / 1.42.
- `.effort-link` — `font-size: 0.85rem`. Was 1.05rem. Cormorant Garamond serif + 700 weight + underline preserved so the title still reads as a heading-shaped link, mirroring `.p-name-link`'s treatment in Builds (also Cormorant Garamond 700 underlined at the body size).
- `.effort-amount` — `font-size: 0.85rem`. Was 0.84rem (cosmetic). Weight 500 preserved so amounts read as metadata, not body prose.
- `.effort-desc` — `font-size: 0.85rem`; `line-height: 1.48`; **`opacity: 0.72` removed**. The opacity drop was the main contributor to the "Community feels smaller" perception — same content type at full visibility now.
- New rule `.panel--black .content-inner a { font-size: 0.85rem; line-height: 1.48 }` parallel to the existing `.panel--yellow .content-inner a`. Required because the generic `.content-inner a` selector at line 422 sets 0.94rem and beats `.effort-link` on specificity (0,1,1 vs 0,1,0); without this rule, the Community panel would inherit 0.94rem regardless of what I set on `.effort-link`. Mirrors how Builds body-link sizing is enforced.

## Why no HTML restructure
The issue's "out of scope" section rules out structural changes. The `.effort-list` / `.effort-link` / `.effort-amount` / `.effort-desc` selectors are used only on the homepage Community panel (verified via grep across `src/`), so the lighter CSS-override approach was sufficient and there's no cross-page regression risk.

## Testing
- [x] Tests pass — CSS-only edit; no schema/route/build/test/HTML touched. The existing Playwright responsive tests cover homepage layout; this change makes Community typography match Builds, not regress against any layout invariant.
- [x] Build succeeds — verified locally; HMR applied each edit cleanly.
- [x] Manual verification — programmatic style measurement at 1440×900 (Mondrian focus state) and 375×812 (mobile stack). All four affected elements render at 13.6px / line-height 20.13 / opacity 1 — exact match to Builds.

## Self-Review
- [x] Correctness: each acceptance criterion in #339 has a measured 13.6px-vs-13.6px confirmation above. No regression on the SCOPE / STACK ribbons (already matched, untouched here). No mobile regression — mobile stack already inherits body size from the panel rules I changed.
- [x] Regression risk: low. Selectors are scoped to `.panel--black` / `.effort-*`, used only on the homepage. The new `.panel--black .content-inner a` rule is a parallel of an existing `.panel--yellow` pattern; it can't affect non-homepage pages because `.panel--black` only exists on the homepage.
- [x] Style and conventions: matches the existing comment-and-rule pattern in this section (#330 / #339 issue numbers in inline comments). Item-title font-family stays Cormorant Garamond serif to mirror Builds' .p-name-link treatment.
- [x] Test coverage: not applicable (typography-only).
- [x] Security and dependency hygiene: zero dependency changes.

## Review path
+20 / −7 in 1 file. Sub-threshold; no protected paths. Internal review only — merge as nathanjohnpayne after CodeRabbit clears.

### DIFF
--- a/src/styles/global.css
+++ b/src/styles/global.css
@@ -871,10 +871,16 @@ a:focus-visible .link-arrow,
   padding-left: 0.6rem;
 }
 
+/* Community section item typography matches the Builds section's body
+   typography (#339). Item titles, amounts, and descriptions all sit at
+   the same body-text size as the Builds project list — same content
+   type → same typography across the site. The font-family stays
+   Cormorant Garamond serif on .effort-link to mirror Builds' .p-name
+   serif treatment for item titles. */
 .effort-link {
   color: inherit;
   font-family: "Cormorant Garamond", Garamond, serif;
-  font-size: 1.05rem;
+  font-size: 0.85rem;
   font-weight: 700;
   line-height: 1.16;
   text-decoration: underline;
@@ -885,7 +891,7 @@ a:focus-visible .link-arrow,
 
 .effort-amount {
   margin-top: 0.12rem;
-  font-size: 0.84rem;
+  font-size: 0.85rem;
   font-weight: 500;
   letter-spacing: 0.01em;
   line-height: 1.2;
@@ -893,9 +899,8 @@ a:focus-visible .link-arrow,
 
 .effort-desc {
   margin-top: 0.10rem;
-  font-size: 0.84rem;
-  line-height: 1.38;
-  opacity: 0.72;
+  font-size: 0.85rem;
+  line-height: 1.48;
 }
 
 .panel--black .content-inner {
@@ -911,8 +916,16 @@ a:focus-visible .link-arrow,
 
 .panel--black .content-inner > p {
   margin-bottom: var(--su);
-  font-size: 0.95rem;
-  line-height: 1.42;
+  font-size: 0.85rem;
+  line-height: 1.48;
+}
+
+/* Parallel to .panel--yellow .content-inner a — beats the generic
+   .content-inner a (0.94rem) rule on specificity so Community item-link
+   text matches Builds body-link size. (#339) */
+.panel--black .content-inner a {
+  font-size: 0.85rem;
+  line-height: 1.48;
 }
 
 .panel--red {


nathanpayne-claude (commented):
Reviewer: nathanpayne-claude (`--comment` per REVIEW_POLICY.md self-review rule).

+20/-7 in src/styles/global.css. Pure typography harmonization — no schema, route, build, or HTML structure touched.

Empirically verified at 1440×900 and 375×812. After the change, Builds and Community body text both render at 13.6px / line-height 20.13 / opacity 1 across all four equivalent elements (intro, item title, item amount, item description). The 0.72→1 opacity removal on `.effort-desc` was the largest single contributor to the "Community feels smaller" perception.

Specificity worth flagging: the new `.panel--black .content-inner a` rule is required because the generic `.content-inner a` (0,1,1) beats `.effort-link` (0,1,0) and would override the 0.85rem set there. Mirrors the existing `.panel--yellow .content-inner a` rule which solves the same problem on the Builds side.

No cross-page risk: `.effort-*` selectors are only used on the Community panel (verified via grep). `.panel--black .content-inner a` is also Community-only (panel--black is the homepage Community panel; doesn't appear elsewhere).

Cleared. Awaiting CodeRabbit, then merge.

PR #235, opened 2026-04-18
mux(phase-2): wire PUBLIC_MUX_ENV_KEY + metadata into MuxPlayer

nathanjohnpayne:
## Summary

- Adds [`.env.example`](.env.example) documenting `PUBLIC_MUX_ENV_KEY` — the client-safe Mux Data env key.
- In [src/layouts/ProjectLayout.astro](src/layouts/ProjectLayout.astro), reads `import.meta.env.PUBLIC_MUX_ENV_KEY` and passes three attributes to `<mux-player>`:
  - `env-key={muxEnvKey}` — when set, telemetry goes to Mux Data; when unset, the attribute is omitted entirely and the player emits nothing.
  - `metadata-video-title={headline}` — the project title.
  - `metadata-video-id={slug}` — the project slug (e.g. `swipe-watch`).
- Adds a `slug` prop to `ProjectLayout` and wires `data.slug` through from [src/pages/projects/[slug].astro](src/pages/projects/[slug].astro).
- Adds `!.env.example` to [.gitignore](.gitignore) so the template file clears the existing `.env.*` ignore rule.

Closes [#216](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/216) and [#217](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/217).

Part of [MUX Video Integration — Phase 2](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/215). The remaining sub-issues — provisioning the key in 1Password + Firebase ([#218](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/218)) and QA that views register in Mux Data ([#219](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/219)) — are blocked on human action and handled separately.

## Graceful-degrade behavior

With `PUBLIC_MUX_ENV_KEY` **unset** (dev preview, local machines without `.env.local`):

```html
<mux-player
  playback-id="wNCRY97981o2uDAJrJ3ExPeK379yldRRFJgUIgSYz00k"
  autoplay="muted" muted loop playsinline
  metadata-video-title="Swipe Watch"
  metadata-video-id="swipe-watch"
  …
/>
```

No `env-key` attribute → no analytics events → no console errors → player works identically.

With `PUBLIC_MUX_ENV_KEY` set (post-provisioning in Firebase):

```html
<mux-player
  env-key="XXXXXXXX"
  playback-id="…"
  metadata-video-title="Swipe Watch"
[...]

PR #343, opened 2026-05-13
fix(home): feature-detect requestAnimationFrame in measureContentHeights (#342 → #315 P3)

nathanjohnpayne:
Authoring-Agent: claude

## Summary

Wrap the `requestAnimationFrame(...)` call inside `measureContentHeights()` in [src/pages/index.astro:588](https://github.com/nathanjohnpayne/nathanpaynedotcom/blob/main/src/pages/index.astro#L588) with a `typeof requestAnimationFrame === 'function'` check, matching the surrounding feature-detection style (the same function already feature-detects `document.fonts.ready` at line 594). When rAF is unavailable, fall back to a direct synchronous removal of `is-measuring`.

Resolves the #315 P3 finding from the [#342 validation pass](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/342#issuecomment-4444682862).

## Self-Review

**Correctness.** rAF is universally supported in modern browsers; this guard only matters in unusual SSR/headless environments where rAF is absent. The fallback path executes the same effect (`grid.classList.remove('is-measuring')`) synchronously, which is correct behavior when no animation frame is available to defer to.

**Regression risk.** Minimal. The behavior change is *additive*: identical in the common path (rAF present), and graceful in the previously-broken path (rAF absent → no-op cleanup of the measure-mode class).

**Style.** Matches `index.astro:594`'s feature-detect pattern (`typeof document !== 'undefined' && document.fonts && document.fonts.ready`).

**Test coverage.** No unit tests exercise this branch (the rAF-absent path is an edge case). All 143 vitest tests + 42 Playwright tests pass on this branch.

**Security / dependency hygiene.** Pure JS edit. No new dependencies.

## Test plan

- [x] `npm test` — 143/143 vitest tests pass.
- [x] `npx playwright test tests/responsive/mondrian-rebalance-animation.spec.ts` — 9/9 desktop tests pass; iPad variants correctly skipped (mobile stack mode short-circuits the measure pass).
- [x] Visual: no observable change in modern browsers (rAF is always present).

Note: 3 other items from issue #342 (#254 yellow panel clamp, #306 panel labels w/o JS, #235 Mux env-key) were re-classified during implementation: #306 + #235 are ALREADY-FIXED on closer inspection (the original validation pass missed the `<noscript>` fallback in BaseLayout.astro and the removal of `env-key` entirely from `ProjectMuxPlayer.astro`); #254 has a documented design rationale in the original PR body that conflicts with the reviewer's blanket "use clamp()" complaint — needs a design call, not a code-only fix. A consolidation comment will land on #342 after the remaining PRs.

### DIFF
--- a/src/pages/index.astro
+++ b/src/pages/index.astro
@@ -585,7 +585,11 @@ const homepageJsonLd = {
           if (restoredVisible) restoredActive.classList.add('is-content-visible');
         }
         void grid.offsetHeight;
-        requestAnimationFrame(() => grid.classList.remove('is-measuring'));
+        if (typeof requestAnimationFrame === 'function') {
+          requestAnimationFrame(() => grid.classList.remove('is-measuring'));
+        } else {
+          grid.classList.remove('is-measuring');
+        }
       }
 
       // Run after fonts load (heights depend on Cormorant Garamond's


nathanpayne-claude (commented):

**Correctness.** The guard pattern is correct. `typeof requestAnimationFrame === 'function'` is the right check — it tolerates both `undefined` and the case where a non-function value is bound to the name (rare but possible in some testing harnesses).

**Symmetry with the rest of the file.** Matches the feature-detect style at line 594 (`document.fonts.ready`) and 144 / 777 in `BaseLayout.astro`. Reads consistently.

**Fallback semantics.** The synchronous `grid.classList.remove('is-measuring')` is the right fallback. The original rAF call's only purpose is to push the class removal one frame past the layout reflow on line 587 (`void grid.offsetHeight`); on a runtime with no rAF, there's no frame queue to push to, so synchronous removal is the closest equivalent.

**Regression risk.** None observable — the changed branch only fires on rAF-absent runtimes. The 9/9 desktop Mondrian Playwright tests still pass on this branch, confirming the rAF-present path is unchanged.

Nothing to address. Approving via comment per policy (reviewer-identity self-approval is not posted).

PR #193, opened 2026-04-16
content: surface ai-agent-repo-template across the site

nathanjohnpayne:
Authoring-Agent: claude

Closes #186, #187, #188, #189, #190.

## Summary

Tasks 1–5 from `ticket-site-updates-post-agent-approval-workflow.md`, bundled per Nathan's ask for a single PR.

- **Task 1+2** — Add `ai-agent-repo-template` as a first-class project (content file + auto-generated detail page via `[slug].astro`, plus OG card template). Placed first on `/projects/` since it's the infrastructure the others depend on.
- **Task 3** — Rewrite the Projects page intro to lead with the enforcement-system framing and link the new blog post.
- **Task 4** — Add a second link ("How the enforcement system works →") to the homepage Vibe Coding paragraph, separated by a middle dot to match the site's tag-row convention.
- **Task 5** — Audit + one-sentence link from `override.md`, `device-source-of-truth.md`, and `friends-and-family-billing.md` to the new blog post. `swipe-watch.md` skipped because it has no workflow language. Audit details posted on #190.

## GitHub-social hero auto-refresh

For the hero image on the new detail page, this PR introduces an opt-in pattern:

```yaml
heroRefresh: "github-social"
```

in project frontmatter. A `prebuild` hook (`scripts/refresh-hero-images.mjs`) resolves the repo's `og:image` meta tag and downloads the 1200×600 social preview PNG to `screenshotSrc` before Astro builds. If the fetch fails (network, rate-limit, auth, etc.), the script logs a warning and the existing on-disk image stays, so the build never breaks on a transient GitHub outage.

## Screenshots

- New `/projects/` index, with the new card first and updated intro copy. Hero shows the live GitHub social card for `nathanjohnpayne/ai_agent_repo_template`.
- Homepage Vibe Coding section now renders both links inline, separated by `·`.

Verified manually in the dev server across all five project pages and the homepage.

## Self-Review

- **Correctness** — Verified: new project content file validates against the extended schema (new optional `heroRefresh` enum); `[slug].astro` picks it up via `getStaticPaths()`; all 134 vitest tests pass including the new project's render suite. No separate `.astro` file needed — the dynamic route already templates every project.
- **Regression risk** — Low. Copy changes are scoped to single strings on `index.astro` (homepage) and `projects/index.astro`. Three project markdown files got one appended sentence each. The prebuild hook is a new dev/deploy step but it is idempotent and fails soft. Existing hero images for Override/DST/FFB/Swipe Watch are not touched.
[...]

PR #64, opened 2026-04-08
Phases 8+9: Update tests and remove legacy files

nathanjohnpayne:
### Phase 8: Tests and CI updates (#43)
- Rewrite all 8 test files to read from `dist/` instead of legacy root paths
- Handle Astro's hashed CSS filenames (`dist/_astro/*.css`)
- Handle inline script extraction (Astro inlines panel JS, tests must mock `matchMedia` first)
- Update sitemap test for Astro's `sitemap-index.xml` format
- Add `tests/no-external-images.test.js` — regression guard for drift item #3
- Add `tests/content-schema.test.js` — validates blog collection schema and frontmatter
- **76 tests across 11 files, all passing**

### Phase 9: Cleanup (#44, #48)
- Delete legacy files: `index.html`, `style.css`, `script.js`, `sitemap.xml`
- Delete `projects/*/index.html` (4 files)
- Delete `blog/` directory (index, post HTML, images)
- Delete `content/blog/` (moved to `src/content/blog/` in Phase 4)
- Delete `scripts/generate-blog.js` (replaced by Astro + remark/rehype)
- Delete root-level static assets (now in `public/`): `og-image.png`, `og/`, `favicon.svg`, `favicon-32x32.png`, `apple-touch-icon.png`, `robots.txt`
- Update `rules/repo_rules.md` for Astro structure
- Update `.repo-template.yml` to remove deleted directories

Authoring-Agent: claude

Closes #43, closes #44, closes #48
Parent: #33

## Self-Review

**Correctness:** 76 tests pass across 11 files. Build produces 7 pages. All CI enforcement scripts pass. No legacy files remain.

**Regression risk:** High — this is the point of no return. Legacy files are deleted. Rollback path: revert this commit, set `firebase.json` `public` back to `.`, redeploy. All deleted content exists in git history.

**Style:** Test rewrites follow the same patterns (readFileSync from dist/, script stripping, CSS glob).

**Test coverage:** No test reads from legacy paths anymore. New regression tests guard against external image URLs and schema drift.

**Security/dependency hygiene:** No new dependencies. Removed `scripts/generate-blog.js` (no longer needed).

## Release Notes

* **Refactor**
  * Restructured site pages and build system including homepage, blog section, and project showcase
  * Updated repository configuration and structural organization

* **Tests**
  * Updated test infrastructure to reference new build output directories and format
  * Added new validation tests for content schema and external resources

PR #348, opened 2026-05-13
refactor(scripts): swap hand-rolled frontmatter parser for js-yaml (#193, #64)

nathanjohnpayne:
Authoring-Agent: claude

## Summary

The prebuild scripts (`scripts/refresh-hero-images.mjs`, `scripts/refresh-mux-gifs.mjs`) each carried an identical hand-rolled frontmatter parser that read line-by-line `key: value` text and only stripped double quotes:

```js
for (const line of body.split('\n')) {
  const m = line.match(/^(\w+):\s*(.*)$/);
  ...
  if (value.startsWith('"') && value.endsWith('"')) value = value.slice(1, -1);
}
```

Single-quoted strings, inline arrays, and nested objects were silently mangled. Production callers happened to only access flat string fields (`heroRefresh`, `muxPlaybackId`, `slug`, `screenshotSrc`, `githubUrl`) so the bug was not load-bearing — but the brittle parsing made the surface easy to regress whenever a new prebuild consumer touched a richer key.

This PR replaces both copies with a shared helper at `scripts/lib/parse-frontmatter.mjs` that uses `js-yaml`. The dep is already in the project's dep graph transitively (Astro consumes it); we pin a direct `devDependency` here so the import contract is explicit and decoupled from Astro version drift.

Resolves the #193 P1 + #64 nit findings from the [#342 validation pass](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/342#issuecomment-4444682862). The nit is a duplicate of the P1 — both close together.

## Self-Review

**Correctness.** I verified the new parser against every project markdown file under `src/content/projects/` and confirmed it produces the same flat-string field values as before, while now correctly preserving:

- `tags: ["Consumer", "Streaming", ...]` arrays (the old parser stored them as a string)
- nested `metadata:` blocks (the old parser silently dropped them)
- single-quoted strings (the old parser dropped these)
- CRLF line endings (the regex on the closing `\r?\n---` is unchanged)

No production caller accesses `tags` or `metadata` today, so this is forward-protection rather than a behavior change.

**Regression risk.** Low. The new helper is a strict superset of the old parser's contract — every input the old parser handled correctly is still handled correctly, and inputs the old parser mishandled are now handled per YAML 1.1.

**Style.** The shared helper sits alongside the consumers (`scripts/lib/parse-frontmatter.mjs`), matching the existing `scripts/lib/` convention. Both prebuild scripts import via a single named export.

**Test coverage.** Added `tests/parse-frontmatter.test.js` with seven cases covering the forms the old parser mishandled (single-quoted strings, inline arrays, nested objects, CRLF endings, throws on malformed YAML) plus the production-path flat-field forms. Total vitest suite: 151 passed (143 baseline + 8 new).

**Security / dependency hygiene.** Adds `js-yaml ^4.1.1` as a direct `devDependency`. `js-yaml` is the canonical YAML parser for Node, has no transitive runtime deps in its core path, and was already present in `node_modules` via Astro. The pinned range matches the version Astro pulls. Pure dev-time dep — does not affect production bundle.

## Test plan

- [x] `npm test` → 151/151 vitest tests pass.
- [x] Spot-check: parsed `slug`, `heroRefresh`, `muxPlaybackId`, `screenshotSrc`, `githubUrl` for all 6 projects → identical to old parser output.
- [x] Spot-check: `tags` and `metadata` on the same set of files now parse to their real structured shapes (array / object respectively) rather than the string-or-empty shapes the old parser produced.
- [ ] CI lane will re-run `npm test` + the Playwright suite.

Note: this PR does not run the prebuild scripts themselves (those hit the network — GitHub for social images, Mux for GIFs). Their parser swap is exercised by the unit test + the spot-check above; the network behavior is unchanged.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "@astrojs/sitemap": "^3.0.0",
     "@playwright/test": "^1.59.1",
     "astro": "^6.1.0",
+    "js-yaml": "^4.1.1",
     "jsdom": "^29.0.2",
     "unist-util-visit": "^5.1.0",
     "vitest": "^4.1.4"
--- a/scripts/lib/parse-frontmatter.mjs
+++ b/scripts/lib/parse-frontmatter.mjs
@@ -23,34 +23,38 @@
  * explicit and decouple from Astro version drift.
  */
 
-import { load as parseYaml } from 'js-yaml';
+import { load as parseYaml, FAILSAFE_SCHEMA } from 'js-yaml';
 
 const FRONTMATTER_RE = /^---\r?\n([\s\S]*?)\r?\n---/;
 
 /**
- * Top-level primitive coercion. js-yaml correctly types scalars (a bare
- * `123` parses to a JS number, `true` to a boolean, etc.), but the
- * production callers (`refresh-hero-images.mjs`, `refresh-mux-gifs.mjs`)
- * inherited the previous hand-rolled parser's "every value is a string"
- * contract: e.g. `data?.screenshotSrc?.startsWith('/')` would throw a
- * TypeError if the value parsed to a number, and `parseGithubRepo(githubUrl)`
- * calls `.match()` on its input. Coerce top-level primitives back to
- * strings to preserve the old contract, leaving arrays / nested objects
- * untouched (those are new capabilities the old parser silently mangled).
+ * Schema choice: FAILSAFE_SCHEMA, not the default.
  *
- * Codex P2 catch on PR #348.
+ * js-yaml's DEFAULT_SCHEMA resolves implicit tags for numbers, booleans,
+ * null, AND timestamps (ISO-like bare scalars → Date objects). The
+ * production callers in this repo (`refresh-hero-images.mjs`,
+ * `refresh-mux-gifs.mjs`) inherited the previous hand-rolled parser's
+ * "every leaf is a string" contract — e.g.
+ *   data?.screenshotSrc?.startsWith('/')
+ *   parseGithubRepo(githubUrl).match(...)
+ * would throw `TypeError: …is not a function` on a number or Date.
+ *
+ * FAILSAFE_SCHEMA resolves only `!!str`, `!!seq`, `!!map`. Bare scalars
+ * stay as strings ("42", "true", "2026-05-13" remain strings, not
+ * coerced to JS number / boolean / Date). Containers — arrays and
+ * nested objects — are still parsed structurally (the actual #193 +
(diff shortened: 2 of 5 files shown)


nathanpayne-claude (commented):

**Correctness.** The swap is a strict-superset behavior change:
- Every flat-string field the production callers access (`heroRefresh`, `muxPlaybackId`, `slug`, `screenshotSrc`, `githubUrl`) deserializes identically. Verified by walking all 6 project files in `src/content/projects/`.
- `tags` arrays and `metadata:` nested objects now deserialize to their real shapes. No production caller reads either field today, so this is forward-protection.
- The closing-fence regex `^---\r?\n([\s\S]*?)\r?\n---` is unchanged, so the only behavior delta is *what's done with the captured body* — handed to `js-yaml.load` instead of split-by-newline.

**The `throws on malformed YAML` test is the right call.** The prebuild scripts intentionally don't catch around `parseFrontmatter` — a corrupted frontmatter block should fail the build loudly, which is the design intent on `refresh-mux-gifs.mjs` per its module header comment ("Failure semantics are strict by design"). `refresh-hero-images.mjs` has softer failure semantics around the *network* fetch but a corrupt frontmatter file is a different class of failure and the same loud-fail is correct.

**Dep hygiene.** `js-yaml ^4.1.1` is the right pin. It matches what Astro already pulls in transitively, so there's no new node_modules cost. The lockfile isn't tracked (per `.gitignore`), so the explicit `package.json` range is the binding source of truth — `^4.1.1` accepts the patch/minor stream which is appropriate for `js-yaml`'s stability history.

**Test coverage.** The new tests/parse-frontmatter.test.js covers:
- Null on no-frontmatter and on empty input
- {} on whitespace-only frontmatter
- Flat string fields the prebuild scripts depend on
- Single-quoted strings (the old parser silently dropped these)
- Inline arrays (the old parser stringified these)
- Nested objects (the old parser silently dropped these)
- CRLF line endings
- Throws on malformed YAML

That's a tight contract — any regression in the parser surface would trip at least one of these.

**Style.** Shared helper at `scripts/lib/parse-frontmatter.mjs` follows the existing `scripts/lib/` convention. The JSDoc on `parseFrontmatter` documents the contract clearly, including the choice to throw rather than swallow.

nathanpayne-claude:
@codex review

nathanjohnpayne:
@codex review

nathanjohnpayne:
@codex review

PR #349, opened 2026-05-13
fix(scripts): harden preflight + CodeRabbit wait against three #250 findings

nathanjohnpayne:
Authoring-Agent: claude

## Summary

Three small but real correctness fixes in the cowork-credential plumbing, all surfaced by the [#342 validation pass](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/342#issuecomment-4444682862):

### 1. `op-preflight.sh` — require `--agent` in deploy mode (#250 P3)

`--mode deploy` previously allowed an empty `--agent`, but cache paths derive directly from `$AGENT`:

```
$CACHE_DIR/op-preflight-$AGENT.env
$CACHE_DIR/op-preflight-$AGENT-adc.json
```

An empty agent produced shared filenames (`op-preflight-.env`) that collide across invocations and silently mix credentials between cowork sessions. The fix extends the existing review/all `--agent` gate to deploy mode, with the same fail-fast shape (`exit 2` + usage line).

### 2. `op-preflight.sh` — refresh ADC when cache-hit ADC is unusable (#250 P2)

The fast path previously detected a stale ADC via `adc_is_usable`, warned, dropped the ADC exports, and still returned success. A user who had rotated the 1Password ADC since the cache was written would stay degraded until TTL expiry (4 h), with no way to opt back in besides `--refresh`.

Fix: on cache-hit + unusable ADC, `exit 2` to fall through to the full-fetch path. That matches how a missing-ADC file was already handled three lines up — symmetry with the existing `[[ -z … ]] || [[ ! -s … ]] && exit 2` check.

Trade-off documented in the inline comment: when 1Password still holds the same stale ADC (i.e., the rotation didn't happen yet), the slow path itself logs guidance and continues with PATs only — so the user gets *one* biometric prompt rather than waiting out TTL. Self-healing wins.

### 3. `coderabbit-wait.sh` — anchor freshness to GitHub push time (#250 P2)

The freshness anchor (`HEAD_ANCHOR`) previously fell back to `HEAD_COMMITTER_DATE` with only a `head_ref_force_pushed` advance. After a normal (non-force) push of a commit whose committer date predates the push (cherry-pick, rebase with `--committer-date-is-author-date`, etc.), stale CodeRabbit comments created between the committer date and the push time would be incorrectly treated as "for the current HEAD".

Fix: use the PR timeline's `committed` event for `HEAD_SHA`. Its `created_at` field is the time GitHub *received* the commit (push time), independent of the local commit's `committer.date`. The existing `head_ref_force_pushed` advance still runs on top, so force-pushed PRs continue to anchor to the force-push event time.

## Self-Review

**Correctness.** All three changes preserve existing behavior in the happy path and only alter the previously-buggy paths:
- The `--agent`/deploy gate is additive — old well-formed invocations (`--mode deploy --agent claude`) continue unchanged.
- The cache-hit ADC fix re-uses the existing `exit 2` → full-fetch fall-through that already covers the missing-file case. The full-fetch path itself unchanged.
- The push-time anchor is preferred *when present* and is bounded below by `HEAD_COMMITTER_DATE`, then bounded above by `head_ref_force_pushed` — never moves the anchor *backward*.

**Regression risk.** Low.
- `op-preflight.sh --mode deploy` callers who previously relied on omitting `--agent` would now error — verified there are no such callsites in this repo (`grep -rn "preflight.*--mode deploy" | grep -v "agent"` is empty).
- The cache-hit ADC `exit 2` change can produce one additional biometric prompt per session in the edge case where 1Password ADC source is broken (not stale). Documented in the inline comment.
- The push-time anchor fix can only move `HEAD_ANCHOR` *forward* (more recent). The freshness filter is `comment.created_at >= HEAD_ANCHOR`, so moving forward = fewer false positives, never more.

**Style.** Each change includes an inline block comment explaining the rationale + the validation-pass cross-reference (#342 → #250 …). Matches the existing comment density in both files.

**Test coverage.** No vitest/Playwright coverage exists for shell-script behavior; verified manually:

- `bash -n` clean on both files.
- `bash scripts/op-preflight.sh --mode deploy --dry-run` → errors with the new message.
- `bash scripts/op-preflight.sh --mode review --dry-run` → errors with the new message (existing behavior, unchanged).
- `bash scripts/op-preflight.sh --agent claude --mode deploy --dry-run` → dry-run banner prints, no error.
- `npm test` → 143/143 vitest unchanged.

**Security / dependency hygiene.** No new dependencies. Touches no credential storage. The ADC fix actually *improves* security posture by ensuring rotated credentials are picked up sooner.

## Test plan

- [x] `bash -n scripts/op-preflight.sh && bash -n scripts/coderabbit-wait.sh` → clean.
- [x] op-preflight dry-run with each `--mode` and with/without `--agent`.
- [x] `npm test` → 143/143 vitest.
- [ ] First reviewer cycle on this PR will exercise the updated `coderabbit-wait.sh` anchor on a live PR — the anchor source is logged so any regression is visible in CI output.

### DIFF
--- a/scripts/coderabbit-wait.sh
+++ b/scripts/coderabbit-wait.sh
@@ -217,9 +217,27 @@ HEAD_COMMITTER_DATE=$(gh api "repos/$REPO/commits/$HEAD_SHA" --jq '.commit.commi
 # codex-review-request.sh: advance the anchor past any `head_ref_force_pushed`
 # timeline event. See nathanjohnpayne/mergepath#140 round-2 Codex finding
 # (P1, line 270) for the specific exposure this closes.
+#
+# Push-time anchoring (#342 → #250 P2 follow-up):
+#   The PR timeline's `committed` event for HEAD_SHA carries a `created_at`
+#   field that is the time GitHub *received* the commit (i.e. push time),
+#   independent of the local commit's `committer.date`. When that event
+#   exists, prefer it over the local committer date — it is the only
+#   timestamp guaranteed to be monotonic with the moment HEAD became
+#   visible to CodeRabbit. Falls back to committer date when the timeline
+#   does not yet contain a matching `committed` event (rare, but happens
+#   on very fresh pushes that race the timeline indexer).
 HEAD_ANCHOR="$HEAD_COMMITTER_DATE"
 ANCHOR_SOURCE="HEAD committer date"
 TIMELINE_JSON=$(fetch_api_array "repos/$REPO/issues/$PR_NUMBER/timeline" "PR timeline")
+HEAD_PUSH_TIME=$(echo "$TIMELINE_JSON" | jq -r --arg sha "$HEAD_SHA" '
+  [ .[] | select(.event == "committed" and .sha == $sha) | .created_at ]
+  | max // ""
+')
+if [ -n "$HEAD_PUSH_TIME" ] && [[ "$HEAD_PUSH_TIME" > "$HEAD_ANCHOR" ]]; then
+  HEAD_ANCHOR="$HEAD_PUSH_TIME"
+  ANCHOR_SOURCE="committed event @ $HEAD_PUSH_TIME (push time)"
+fi
 LATEST_FORCE_PUSH_TIME=$(echo "$TIMELINE_JSON" | jq -r '
   [ .[] | select(.event == "head_ref_force_pushed") | .created_at ]
   | max // ""
--- a/scripts/op-preflight.sh
+++ b/scripts/op-preflight.sh
@@ -128,10 +128,15 @@ if $PURGE_ALL; then
   exit 0
 fi
 
-if [[ "$MODE" == "review" || "$MODE" == "all" || "$PURGE" == "true" ]] && [[ -z "$AGENT" ]]; then
-  echo "Error: --agent is required for review, all, or --purge mode." >&2
+if [[ "$MODE" == "review" || "$MODE" == "deploy" || "$MODE" == "all" || "$PURGE" == "true" ]] && [[ -z "$AGENT" ]]; then
+  # Deploy mode also requires --agent because the cache file paths below
+  # are derived from $AGENT (op-preflight-$AGENT.env / op-preflight-$AGENT-adc.json).
+  # An empty agent produces shared filenames (`op-preflight-.env`) that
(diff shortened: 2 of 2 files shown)


nathanpayne-claude (commented):

**Correctness.** All three changes hold up under inspection:

- The `--agent`/deploy gate is a strict addition to the existing review/all gate. The exit code change (`exit 1` → `exit 2`) brings the message-emit path in line with the rest of preflight's "argument error" convention (`exit 2` for "caller error / re-fetch needed"). The behavior change is observable only for callers who omitted `--agent` with `--mode deploy` — verified zero such call sites in the repo.
- The cache-hit ADC `exit 2` fix correctly delegates to the slow path, which already has stale-ADC handling at lines 503-507 (warn + drop exports + emit PATs, no error). So the net effect across both paths is now: "1Password ADC genuinely broken" continues to degrade gracefully, "1Password ADC rotated" now self-heals on next preflight call. That's the right asymmetry to introduce — the previous code was symmetric in the wrong direction (degraded forever until TTL).
- The push-time anchor fix is the cleanest single change of the three. The `committed` event's `created_at` is the canonical push-arrival timestamp; relying on it gives us the right anchor floor whether the user did a normal push, a rebase, or a cherry-pick. The fall-through to `HEAD_COMMITTER_DATE` covers the timeline-indexer race window, and the existing `head_ref_force_pushed` advance still wins when present.

**Monotonicity guarantee.** I checked that `HEAD_ANCHOR` can only move forward across the three branches (committer date → committed event time → force-push time). Each subsequent block uses `[[ "$X" > "$HEAD_ANCHOR" ]]` to compare before advancing, so the anchor never regresses. That preserves the invariant the rest of the script (the `comments.created_at >= HEAD_ANCHOR` filter) depends on.

**Style.** The inline comment block on the new anchor logic is appropriately verbose — it documents the bug class, the timeline event used, and the fall-through behavior, in the same shape as the existing force-push advance comment. The `#342 → #250 …` cross-references are consistent across both files and tie back to the validation report.

**Trade-off acknowledgment.** The PR body correctly calls out the "one extra biometric prompt on source-broken ADC" cost of the self-healing fix. That's the right disclosure — anyone reading the diff later will understand why the cache-hit path now diverges from "always emit on cache fresh" semantics.

PR #232, opened 2026-04-18
tooling: add scripts/gh-projects/ for multi-phase Project v2 initiatives

nathanjohnpayne:
## Summary

- Adds `scripts/gh-projects/` — a sourceable helper library plus standalone CLI for driving multi-phase GitHub Projects v2 initiatives.
- `lib.sh` exposes `create_parent`, `create_child`, `link_sub_issue`, `add_to_project`, `set_project_readme`, `ensure_label`, and `prep_body` (placeholder substitution on body files).
- `move-item.sh` moves an issue to a named Status swimlane by discovering the project's node ID, field ID, and option IDs at runtime — works against any Project v2 with a `Status` single-select field.
- `examples/mux-video-integration/` preserves the exact body-file templates and rerunnable `create-issues.sh` driver used to create the MUX initiative's issue tree (#210–#230) — serves as the canonical reference for the next multi-phase initiative.

Closes #231.

## Self-Review

**Correctness.** `lib.sh` encapsulates three subtle behaviors that matter for correctness:

1. `link_sub_issue` uses `gh api -F sub_issue_id=<id>` (typed/integer), not `-f` (string). The REST sub-issues endpoint returns `422 Invalid property` on string input — verified in the session that produced #210/#211.
2. The sub-issue ID passed in is the issue's DB `id` (integer like `4288428629`), not the issue number. `create_child` fetches it via `gh api repos/.../issues/<num> --jq .id`.
3. `move-item.sh` discovers the project node ID via `gh project view --format json` separately from the `Status` field/option IDs (via `gh project field-list`). Tested live against Project #5 — #210 and #211 were moved to "In Progress" using this path.

**Regression risk.** Zero — this is a purely additive directory. No existing scripts, configs, workflows, or runtime code changed. `bash -n` passes on all three shell files.

**Style.** Shell files use `set -euo pipefail` and are idempotent where possible (`ensure_label --force`). Placeholder substitution (`__PARENT_NUM__` etc.) is documented in both `lib.sh` header comments and the README. No emojis in any file per the repo's convention.

**Test coverage.** No automated tests — these are one-shot driver scripts invoked manually once per initiative. The worked example under `examples/mux-video-integration/` is the de-facto integration test; re-running it against a fresh project would reproduce the current issue tree.

[...]

PR #350, opened 2026-05-13
fix(gh-projects): matchline tooling nits from #342 (#292, #232)

nathanjohnpayne:
Authoring-Agent: claude

## Summary

Two small hardening fixes in the gh-projects matchline driver, both from the [#342 validation pass](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/342#issuecomment-4444682862):

### 1. #292 P2 — pick the oldest matching parent issue, not the newest

`find_parent_num()` in `additions/add-plan-refinements.sh` called `gh issue list --search "in:title ..."` without specifying sort order. The CLI default is newest-first, so if a repo accidentally accumulated duplicate parent titles across re-runs of `create-issues.sh`, the script would pick the most recently-created (typically the empty/abandoned duplicate) and re-attach refinement work under the wrong parent.

Add `--sort created --order asc` so we deterministically pick the *original* parent.

### 2. #232 nit — trap-clean auto-created GHP_TMPDIR on exit

`scripts/gh-projects/lib.sh` minted `GHP_TMPDIR="$(mktemp -d)"` unconditionally when the caller hadn't pre-exported one, but never removed it on exit. Re-running `create-issues.sh` repeatedly left a graveyard of `$TMPDIR/tmp.XXXXXX/phase-*-c*.md` files behind.

Track ownership:

- If the script created the dir → set `GHP_TMPDIR_OWNED=1` and register `trap 'rm -rf "$GHP_TMPDIR"' EXIT`.
- If the caller pre-exported the dir → leave its lifecycle to them (some wrappers want to inspect generated bodies after the run).

### Item dropped from this PR's scope

The third item the [#342 validation report](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/342#issuecomment-4444682862) flagged in this cluster — **#263 · pass C3 child number to Phase 1 C7 prep** — is being **reclassified as MOOT** in the follow-up summary. On inspection:

- `scripts/gh-projects/examples/matchline/phase-1/c7.md` only ever contained `__C1_NUM__`, never `__C3_NUM__` (verified via `git log -S "__C3_NUM__" -- scripts/gh-projects/examples/matchline/phase-1/c7.md`).
- The existing call in `create-issues.sh:131` correctly passes one c-arg (`$P1_C6_NUM`).
- The reviewer's description in PR #263 referred to a draft state that didn't ship.

The `__C{1,2,3,4}_NUM__` placeholders in `lib.sh::prep_body` are *positional* slots for the current child's ordered dependency list, not literal "phase's c1/c3" references — verified by tracing the call shape across c5/c8/c9 which all use the same positional convention.

## Self-Review

**Correctness.**
- `--sort created --order asc` is the documented `gh issue list` flag pair. The post-filter `.[] | select(.title == "$title") | head -1` then takes the first (oldest) exact-title match.
- The `trap … EXIT` only registers when the script itself owns the dir; caller-owned dirs are explicitly skipped via the `GHP_TMPDIR_OWNED` sentinel. Bash `trap` is per-shell, so this does not affect anything outside the sourced script's lifetime.

**Regression risk.** Low.
- In the common case (no duplicate parent issues), `--sort created --order asc` produces the same single match as the previous unsorted call — newest-first vs oldest-first is irrelevant when N=1.
- The trap is additive when there's no pre-existing `GHP_TMPDIR`. When there is a pre-existing one, behavior is unchanged.

**Style.** Inline comments on each change explain the rationale and cross-reference `#342 → …`.

**Test coverage.** No automated test framework for the gh-projects shell side; verified by focused bash smoke tests:

- **Script-owned dir:** created, file written, dir removed at exit.
- **Caller-owned dir:** created, file written, dir preserved at exit.

`npm test` unchanged: 143/143 vitest.

**Security / dependency hygiene.** No new dependencies. No credential changes.

## Test plan

- [x] `bash -n scripts/gh-projects/lib.sh && bash -n scripts/gh-projects/examples/matchline/additions/add-plan-refinements.sh` → clean.
- [x] Smoke test the trap-fires path → tmpdir cleaned up at exit.
- [x] Smoke test the caller-owned path → tmpdir preserved at exit.
- [x] `npm test` → 143/143 vitest.

### DIFF
--- a/scripts/gh-projects/examples/matchline/additions/add-plan-refinements.sh
+++ b/scripts/gh-projects/examples/matchline/additions/add-plan-refinements.sh
@@ -33,14 +33,16 @@ source "$SCRIPT_DIR/../../../lib.sh"
 find_parent_num() {
   local title="$1"
   local num
-  # --sort created --order asc picks the OLDEST matching issue if a repo
-  # accidentally accumulated duplicate parent titles across re-runs of
-  # create-issues.sh. The CLI default is newest-first, which would pick
-  # the most recently-created (typically the empty/abandoned duplicate)
-  # and re-attach refinement work under the wrong parent. See #342 → #292.
+  # `sort:created-asc` inside --search picks the OLDEST matching issue if
+  # a repo accidentally accumulated duplicate parent titles across re-runs
+  # of create-issues.sh. (`gh issue list` itself does not accept bare
+  # `--sort`/`--order` flags — sort directives go through the GitHub
+  # issue-search qualifier syntax, per the `gh issue list --help` examples
+  # and Codex P1 catch on PR #350.) The default issue-search ordering is
+  # "best match", which can pick the empty/abandoned duplicate first and
+  # re-attach refinement work under the wrong parent. See #342 → #292.
   num=$(gh issue list --repo "$REPO" --state all --limit 200 \
-    --search "in:title \"$title\"" \
-    --sort created --order asc \
+    --search "in:title \"$title\" sort:created-asc" \
     --json number,title \
     --jq ".[] | select(.title == \"$title\") | .number" \
     | head -1)
--- a/scripts/gh-projects/examples/matchline/additions/add-plan-refinements.sh
+++ b/scripts/gh-projects/examples/matchline/additions/add-plan-refinements.sh
@@ -33,8 +33,14 @@ source "$SCRIPT_DIR/../../../lib.sh"
 find_parent_num() {
   local title="$1"
   local num
+  # --sort created --order asc picks the OLDEST matching issue if a repo
+  # accidentally accumulated duplicate parent titles across re-runs of
+  # create-issues.sh. The CLI default is newest-first, which would pick
+  # the most recently-created (typically the empty/abandoned duplicate)
+  # and re-attach refinement work under the wrong parent. See #342 → #292.
   num=$(gh issue list --repo "$REPO" --state all --limit 200 \
     --search "in:title \"$title\"" \
+    --sort created --order asc \
     --json number,title \
     --jq ".[] | select(.title == \"$title\") | .number" \
     | head -1)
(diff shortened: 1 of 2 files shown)


nathanpayne-claude (commented):

**Correctness.** Both changes hold up.

- `--sort created --order asc` is canonical `gh issue list` syntax, and the subsequent `select(.title == "$title") | head -1` correctly takes the first (oldest) exact-title match. No off-by-one risk because the previous default also took `head -1` from the unsorted set — the ordering is the only thing that changed.
- The trap-ownership pattern is the right shape. `GHP_TMPDIR_OWNED=1` is local to this sourced library; consumers don't need to think about it. The trap is registered immediately after `mktemp -d`, so an early-exit between `mktemp` and the rest of the script still triggers cleanup.

**The reclassification of #263 C3 → MOOT is correct.** I cross-checked: `git log -S "__C3_NUM__" -- scripts/gh-projects/examples/matchline/phase-1/c7.md` returns empty, and the current call site at create-issues.sh:131 passes the right `$P1_C6_NUM` for c7's only positional placeholder (`__C1_NUM__`, semantically "the first dependency of this child" — c6 produces the approved-matches input the generation ticket needs). The positional-vs-literal placeholder convention is consistent across c5/c8/c9. Including the reclassification in this PR's body (rather than silently dropping) is the right disclosure.

**Trap shape.** I considered whether `trap 'rm -rf "$GHP_TMPDIR"' EXIT` could fire in a way that surprises a caller — e.g., if `lib.sh` is sourced from another library that's then sourced from a wrapper. The answer is no: bash `trap` is per-shell, so the trap registered here only fires when *this* shell exits. A wrapper that re-uses `GHP_TMPDIR` across multiple `source lib.sh` calls in the same shell will hit the `GHP_TMPDIR_OWNED` short-circuit on the second source and leave the dir alone.

**Smoke tests demonstrate both branches.** The PR body shows the script-owned path cleans up at exit, and the caller-owned path preserves the dir. That's the contract the trap should implement, and it does.

**Style.** Inline comments on each change with the `#342 → …` cross-ref are consistent with the comment density on the same files.

nathanpayne-claude:
@codex review

nathanjohnpayne:
@codex review

PR #358, opened 2026-05-13
test: lock in cross-platform path conversion for build scripts (#355)

### LINKED ISSUE #355: design: cross-platform path-conversion coverage (Windows CI lane vs. targeted test) (deferred from #342)
## Background

From the [#342 validation pass](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/342) (CodeRabbit nit #9): no Windows CI lane exists to catch the regression class of [#174](https://github.com/nathanjohnpayne/nathanpaynedotcom/pull/174), where a missing `fileURLToPath` call broke OG-image generation on Windows.

Original reviewer comment: https://github.com/nathanjohnpayne/nathanpaynedotcom/pull/174#discussion_r3081943260

## Current state

- The specific bug PR #174 fixed (using a `file://` URL where a filesystem path was expected) is mitigated in `scripts/refresh-hero-images.mjs:26` via `fileURLToPath(import.meta.url)`. The cross-platform contract is correct on `main`.
- No `.github/workflows/` job runs on `windows-latest`. The full CI matrix is Linux-only.

## Reviewer's concern

The regression class is *path conversion correctness on Windows*. Today's code is correct, but nothing enforces the contract continuously. The next developer who writes
```js
const filePath = import.meta.url.replace('file://', '');
```
on a Linux machine ships a Windows-broken build with no signal.

## Why this is deferred

Adding a Windows lane is a **roadmap call**, not a discrete code fix:

- **Cost.** Windows runners are slower and more expensive on GitHub-hosted CI. Astro + Playwright on Windows has its own setup quirks (path separators, line endings, slower playwright browser install).
- **Coverage value.** This site has zero Windows-targeted features. The only meaningful Windows risk is in build-time scripts that handle paths — a small, slow-moving surface. Adding a full Windows lane to catch a 1-file regression class is overkill; a *targeted* integration test exercising path conversion on a Linux runner would catch most regressions at far lower cost.
- **Alternative.** A single Linux-runner test that:
  1. takes `new URL('./fixture.txt', import.meta.url)`,
  2. runs it through `fileURLToPath`,
  3. asserts the result has no `file:` prefix and `existsSync` returns true,

  catches the regression class without a Windows lane. Lighter, but still preventive.

## Suggested next steps

Pick one of:
- **Add a targeted unit test** asserting the `fileURLToPath` contract on the build-script path-handling surface. Lightest, catches the specific bug class. Probably ~30 min of work.
- **Add a Windows CI lane** that runs `npm run build` + the vitest suite. Heaviest, catches everything Windows-flavored but expensive on every PR.
- **Document and close.** Accept the residual risk; the bug class is rare enough that catching it in code review (or via a one-off Windows smoke when build scripts change) is acceptable.

## References

- [#342 validation report](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/342#issuecomment-4444682862) — Nitpicks #9
- PR #174 — the original fix
- `scripts/refresh-hero-images.mjs:26` — the cross-platform pattern this protects

nathanjohnpayne:
Authoring-Agent: claude

## Summary

Lock in the cross-platform path-conversion contract for the prebuild scripts via a targeted unit test, per the design discussion in [#355](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/355).

The reviewer concern from [PR #174](https://github.com/nathanjohnpayne/nathanpaynedotcom/pull/174#discussion_r3081943260) — that no CI catches the next regression in the `import.meta.url` → filesystem path conversion class — is addressed without paying the cost of a Windows CI matrix on a static site with no Windows-targeted features.

## What the test covers

`tests/path-conversion.test.js` adds two assertions:

1. **Platform contract.** `fileURLToPath(import.meta.url)` on the test file itself yields a string with no `file:` prefix that `existsSync()` accepts. Catches a Node runtime / CI environment regression in the API behavior the build scripts depend on.

2. **Source-shape contract.** Each of `scripts/refresh-hero-images.mjs` and `scripts/refresh-mux-gifs.mjs` (via `it.each`) is asserted to:
   - contain the `fileURLToPath(import.meta.url)` pattern, and
   - NOT contain the buggy `import.meta.url.replace(` pattern.

   Fragile to intentional refactor — that's the point. If someone refactors back into string replacement on a Mac/Linux dev machine, this test fails on every CI lane the repo already runs, so the Windows-only regression class can't ship.

## Self-Review

**Correctness.** The regexes are scoped:
- `GOOD_FILEURL_CALL = /fileURLToPath\s*\(\s*import\.meta\.url\s*\)/` — matches the canonical Node idiom with whitespace tolerance.
- `BUGGY_REPLACE = /import\.meta\.url\s*\.\s*replace\s*\(/` — narrower than a bare `.replace(` substring search; matches only call sites *on* `import.meta.url`, not unrelated `String.prototype.replace` usage elsewhere.

The platform-contract test uses `import.meta.url` of the test file itself, so it's self-validating — if the assertion fails it means Node's `url` module broke, which is a runtime problem worth knowing about.

**Regression risk.** Pure test addition. No production code touched. The source-shape regex correctly recognizes both scripts on current `main` (verified — both scripts use `dirname(fileURLToPath(import.meta.url))` at the top).

**Style.** Header comment explains the bug class, the design choice (targeted test over Windows lane), and links to #355 for the discussion. Two-assertion shape matches the rest of `tests/`.

**Test coverage.** Vitest goes from 159 → 162 (1 platform-contract + 2 source-shape, one per script).

**Security / dependency hygiene.** No new dependencies. The test imports only from `node:` builtins and the production scripts' source.

## Test plan

- [x] `npm test` → 162/162 vitest.
- [x] Manual mental-walkthrough of the negative case: introducing `import.meta.url.replace(...)` in either build script would trip the per-script `it.each` assertion with a script-named failure message.

## Why not Windows CI?

Discussed at length in [#355](https://github.com/nathanjohnpayne/nathanpaynedotcom/issues/355). Short version: this site has zero Windows-targeted features, so a full Windows CI lane would pay slow + expensive runner cost on every PR to protect a one-file regression class. A targeted unit test catches the same regression class for ~30 lines of test code and runs on the existing Linux lanes.

If the site ever needs to *build* on Windows for real (contributors, Windows-based agents, etc.), revisit the Windows-lane option then. Until then this is the proportional response.

Closes #355.

### DIFF
--- a/tests/path-conversion.test.js
+++ b/tests/path-conversion.test.js
@@ -1,0 +1,80 @@
+import { describe, it, expect } from 'vitest';
+import { fileURLToPath } from 'node:url';
+import { readFileSync, existsSync } from 'node:fs';
+import { resolve } from 'node:path';
+
+/**
+ * path-conversion.test.js — lock in the cross-platform contract for
+ * `import.meta.url` → filesystem-path conversion in build scripts.
+ *
+ * PR #174 fixed a Windows-only break where a build script used
+ *
+ *     const filePath = import.meta.url.replace('file://', '');
+ *
+ * to convert a module URL to a path. On POSIX that almost works
+ * (modulo a leading slash); on Windows it produces invalid paths like
+ * `/C:/Users/...` that no filesystem call accepts.
+ *
+ * The fix is to use `fileURLToPath(import.meta.url)` from `node:url`,
+ * which handles both platforms correctly. This test locks that
+ * contract in two ways:
+ *
+ *   1. **Platform contract** — assert that `fileURLToPath` on
+ *      `import.meta.url` (this test file itself) yields a string that
+ *      has no scheme prefix and that `existsSync` accepts. Catches a
+ *      regression in the Node runtime / build environment.
+ *
+ *   2. **Source-shape contract** — assert that the production build
+ *      scripts (`refresh-hero-images.mjs`, `refresh-mux-gifs.mjs`)
+ *      both contain the `fileURLToPath(import.meta.url)` pattern and
+ *      do NOT contain the buggy `import.meta.url.replace(` pattern.
+ *      Fragile to *intentional* refactor — that's the point. If someone
+ *      ever refactors back into a string replacement on a Mac/Linux
+ *      machine, this test fails on every CI lane the repo runs.
+ *
+ * See issue #355 for the design conversation that led to choosing
+ * this targeted test over a full Windows CI lane.
+ */
+
+const REPO_ROOT = resolve(fileURLToPath(import.meta.url), '..', '..');
+
+const BUILD_SCRIPTS = [
+  'scripts/refresh-hero-images.mjs',
+  'scripts/refresh-mux-gifs.mjs',
+];
+
+// The buggy pattern. Match the call form, not the literal substring —
+// `import.meta.url.replace(...)` with any args is suspicious in the
+// context of these scripts. `String.prototype.replace` exists for
+// real reasons elsewhere, so this regex is scoped to call sites on
+// `import.meta.url` specifically.
+const BUGGY_REPLACE = /import\.meta\.url\s*\.\s*replace\s*\(/;
+
+// The good pattern.
+const GOOD_FILEURL_CALL = /fileURLToPath\s*\(\s*import\.meta\.url\s*\)/;
+
+describe('cross-platform path conversion (#174 / #355)', () => {
+  it('fileURLToPath(import.meta.url) yields a usable filesystem path', () => {
+    const path = fileURLToPath(import.meta.url);
+    expect(path).not.toMatch(/^file:/);
+    expect(path).not.toContain('://');
+    // The path resolves to this very test file, which obviously exists.
+    expect(existsSync(path)).toBe(true);
+  });
+
+  it.each(BUILD_SCRIPTS)(
+    '%s converts import.meta.url via fileURLToPath, not string replacement',
+    (relativeScriptPath) => {
+      const fullPath = resolve(REPO_ROOT, relativeScriptPath);
+      const source = readFileSync(fullPath, 'utf-8');
+      expect(
+        GOOD_FILEURL_CALL.test(source),
+        `${relativeScriptPath} should call fileURLToPath(import.meta.url) — the cross-platform path-conversion idiom`,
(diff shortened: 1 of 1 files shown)


nathanpayne-claude (commented):

**Correctness.** Both assertions are tightly scoped.

The platform-contract test is self-validating: it asserts `fileURLToPath(import.meta.url)` on this test file yields a path that `existsSync()` accepts. If the assertion ever fails, the test file has to exist (the test ran), so the failure mode is "Node's `url` module is broken" — a runtime problem worth knowing about.

The source-shape regexes are correctly narrow:
- `/fileURLToPath\s*\(\s*import\.meta\.url\s*\)/` matches the canonical idiom with whitespace tolerance.
- `/import\.meta\.url\s*\.\s*replace\s*\(/` is *scoped to the call site on `import.meta.url`*, not a bare `.replace(` substring. So unrelated `String.prototype.replace` usage elsewhere in the build scripts (e.g., URL parameter rewriting) won't false-positive.

**`it.each` shape is the right call** — running the two scripts as separate parametrized cases means a regression in one produces a single, specific failure with the script name in the message, not a combined assertion that hides which script broke.

**Fragility is intentional and proportionate.** The reviewer's concern in #355 was that nothing enforces the contract continuously. Fragility-to-refactor is exactly the property that gives this test its value — if someone moves the path-resolution logic somewhere else, the test loudly says "you changed the place this contract lives, update the test." That's better than the test silently passing while the contract erodes.

**Coverage delta.** Vitest 159 → 162 (one platform + two source-shape via `it.each`). Confirmed locally.

**Style.** Header comment explains the bug class, the design choice, and cross-references #355. Matches the header-comment density of `tests/parse-frontmatter.test.js` and `tests/reading-time.test.js`.

**Mental verification of the negative case.** If a future change introduces `const filePath = import.meta.url.replace('file://', '')` in either build script, the per-script `it.each` assertion trips with the script name in the failure message. If the change moves to a *third* build script that this test doesn't enumerate, the test won't catch it — but no test can catch every yet-to-exist file, and the `BUILD_SCRIPTS` constant is right at the top of the file where adding new entries is a one-line change.

nathanjohnpayne:
@codex review

PR #546, opened 2026-06-18
[claude] Improve posthog-integration-skill docs (paths, identify guidance, layout)

nathanjohnpayne:
## Summary

Documentation-only updates to the PostHog Astro integration skill under `docs/posthog-integration-skill/` (reference material, not part of the site build, so no runtime or user-facing impact):

- Fix the `SKILL.md` step-list to point at the `references/` subdirectory where the step files actually live.
- Harden the identify-user guidance to use a stable canonical user ID rather than raw username or form input.
- Fix the Astro layout snippet in `astro.md` to wrap content in html/head/body/slot.

Note: the live site never calls `posthog.identify()` (anonymous capture events only, no auth), so the identify guidance is a docs improvement, not a fix for this site.

Authoring-Agent: claude

## Self-Review

- **Correctness**: Edits confined to `docs/posthog-integration-skill/`. Confirmed via grep that nothing in `src/` or `astro.config` references this directory, so it is not built or shipped.
- **Regression risk**: None to the site — documentation/skill-reference only.
- **Style**: Consistent with the surrounding skill docs.
- **Test coverage**: N/A (docs).
- **Security/dependency hygiene**: Strengthens identify guidance (avoid raw input). No dependencies, secrets, or config touched. Not a protected path; under-threshold.

### DIFF
--- a/docs/posthog-integration-skill/SKILL.md
+++ b/docs/posthog-integration-skill/SKILL.md
@@ -14,10 +14,10 @@ This skill helps you add PostHog analytics to Astro (Static) applications.
 
 Follow these steps in order to complete the integration:
 
-1. `basic-integration-1.0-begin.md` - PostHog Setup - Begin ← **Start here**
-2. `basic-integration-1.1-edit.md` - PostHog Setup - Edit
-3. `basic-integration-1.2-revise.md` - PostHog Setup - Revise
-4. `basic-integration-1.3-conclude.md` - PostHog Setup - Conclusion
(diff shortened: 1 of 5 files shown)


nathanpayne-claude (approved):
Approving: documentation-only updates to the posthog-integration-skill (not built into the site). Reference-path fixes, identify-user guidance hardening, and an Astro layout fix. CodeRabbit clean, CI green, under-threshold and non-protected. Approving.

PR #556, opened 2026-06-24
[claude] Repo-local accuracy nits: is:inline doc, reduced-motion test, resume slugs

### LINKED ISSUE #552: Repo-local accuracy nits: is:inline doc explanation + flaky reduced-motion animation test
Two low-severity repo-local correctness cleanups still apply on origin/main.

- `docs/posthog-integration-skill/references/EXAMPLE.md:95` — "The `is:inline` directive is required to prevent TypeScript errors about `window.posthog`." is technically wrong: `is:inline` tells Astro not to process/bundle the script; it does not suppress TS errors. (The same file already states the correct rationale at L317: "Uses is:inline to prevent Astro from processing the script" — so the doc is internally inconsistent.) Correct the L95 explanation. (source: https://github.com/nathanjohnpayne/nathanpaynedotcom/pull/513#discussion_r3408793079)
- `tests/responsive/mondrian-rebalance-animation.spec.ts:349` — the reduced-motion assertion still gates on a wall-clock ceiling `expect(elapsed, 'open completes in <200ms under reduced-motion').toBeLessThan(200)`. #337 only moved the measurement in-page (performance.now) but kept the timing threshold, so a slow/loaded CI runner can still flake. Replace the wall-clock gate with a deterministic reduced-motion check (e.g. assert the motion CSS resolves to ~0s, or that the `is-content-visible` class flips synchronously under reduced-motion). (source: https://github.com/nathanjohnpayne/nathanpaynedotcom/pull/337#discussion_r3214163409)
- _(minor, folded from a low-value ambiguous finding)_ `src/components/resume/WritingSection.astro` hardcodes a `const essays = [...]` list of `{href, title}` rather than deriving from the blog content collection; confirm the two slugs exist (and consider deriving them so they can't silently 404). (source: https://github.com/nathanjohnpayne/nathanpaynedotcom/pull/400#discussion_r3338677785)

---
_Filed by the unresolved-feedback validation pass for [#524](https://github.com/nathanjohnpayne/mergepath/issues/524) (2026-06-22 sweep); classification rubric in [#236](https://github.com/nathanjohnpayne/mergepath/issues/236)._

nathanjohnpayne:
Authoring-Agent: claude

Closes #552.

Three low-severity repo-local correctness cleanups flagged by the unresolved-feedback validation pass.

### 1. is:inline doc nit (docs/posthog-integration-skill/references/EXAMPLE.md)
L95 claimed the `is:inline` directive prevents TypeScript errors about `window.posthog`, which is incorrect and inconsistent with the correct rationale already stated at L317. Reworded so it states that is:inline tells Astro not to process or bundle the script.

### 2. Deflake the reduced-motion test (tests/responsive/mondrian-rebalance-animation.spec.ts)
The test gated on a wall-clock ceiling (elapsed < 200ms). #337 moved the measurement in-page but kept the threshold, so a slow or loaded CI runner can still flake. Replaced with two deterministic checks: the grid transition-duration computes to 0s under reduced-motion (a computed-style read, no timing); and content reveal lands within a small frame budget (frames, not milliseconds), which a loaded runner cannot flake because a zeroed-timer reveal still resolves in a handful of frames whereas the un-short-circuited 460ms path needs ~28.

### 3. Validate resume essay slugs at build time (src/components/resume/WritingSection.astro)
The selected-essays list hardcoded title plus href. The display titles stay curated (one is a deliberately shortened form of its post title), but the ids are now validated against the published blog collection at build time and the hrefs derive from them, so a renamed or unpublished post fails the build instead of silently shipping a dead link.

## Self-Review

- **Correctness:** Doc reword matches the L317 statement. The resume change preserves rendered output exactly (same /blog/<id>/ hrefs, unchanged titles); all three ids exist in the published collection so the build does not throw.
- **Regression risk:** Low. The doc change is prose-only; the test change is stricter and deterministic, not weaker; the resume change adds a build-time guard with identical output and fails loudly only if a slug is wrong (the intended behavior).
- **Style:** eslint clean on the changed .astro and .ts files; astro check reports 0 errors.
- **Test coverage:** Ran the full spec on the Desktop 1440 project, all 9 tests pass including the rewritten reduced-motion test. Ran astro build, the resume page renders and slug validation passes.
- **Security/dependency hygiene:** No dependency, credential, or workflow changes.

### DIFF
--- a/docs/posthog-integration-skill/references/EXAMPLE.md
+++ b/docs/posthog-integration-skill/references/EXAMPLE.md
@@ -92,7 +92,7 @@ The PostHog snippet is included as an inline script to prevent Astro from proces
 </script>
 ```
 
-The `is:inline` directive is required to prevent TypeScript errors about `window.posthog`.
+The `is:inline` directive tells Astro not to process or bundle the script, so the PostHog snippet is emitted into the page exactly as written.
 
 ### User identification (`src/pages/index.astro`)
 
--- a/src/components/resume/WritingSection.astro
+++ b/src/components/resume/WritingSection.astro
@@ -5,21 +5,41 @@
  * selected essays, using the site's → arrow convention.
  * Section/heading scaffolding via the shared ResumeSection wrapper (#471).
  */
+import { getCollection } from 'astro:content';
 import ResumeSection from './ResumeSection.astro';
+
+// Curated, ordered selection. Display titles are intentionally editorial
+// (e.g. "The HTML Mock-up Is the Spec" is a deliberately shortened form of
+// that post's longer frontmatter title), so they live here rather than being
+// derived from the collection. The ids ARE validated against the published
+// blog collection below, so a renamed or unpublished post fails the build
+// instead of silently shipping a dead link (#552).
 const essays = [
   {
     title: 'Agent Approval Workflow and the Genesis of Mergepath',
-    href: '/blog/agent-approval-workflow-genesis-of-mergepath/',
+    id: 'agent-approval-workflow-genesis-of-mergepath',
   },
   {
     title: 'Six PRs, One Bug: What AI Agents Actually Get Wrong',
-    href: '/blog/six-prs-one-bug-agent-failure-modes/',
+    id: 'six-prs-one-bug-agent-failure-modes',
   },
   {
     title: 'The HTML Mock-up Is the Spec',
-    href: '/blog/html-mockups-as-spec/',
+    id: 'html-mockups-as-spec',
   },
 ];
+
+const publishedBlogIds = new Set(
+  (await getCollection('blog', ({ data }) => !data.draft)).map((post) => post.id),
+);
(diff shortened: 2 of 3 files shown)


nathanpayne-claude (approved):
Reviewed as nathanpayne-claude. (1) EXAMPLE.md L95 reworded to match the correct is:inline rationale at L317. (2) The reduced-motion test now uses two deterministic checks (grid transition-duration is 0s, plus a frame-budget content reveal) instead of the flaky wall-clock ceiling; verified all 9 tests pass on the Desktop 1440 project. (3) WritingSection keeps curated titles but validates essay ids against the published blog collection at build time and derives hrefs, so a renamed or unpublished post fails the build; verified via astro build that rendered output is unchanged. CodeRabbit reported no actionable comments. Under-threshold (67 lines changed, no protected paths). Approving.

